Issue #5, opened 2016-01-15
Can't enter multiple NS records


mortzu:
If I add a second NS entry for the zone it removes the first one

ivanfilippov:
This also affects MX records.

ivanfilippov:
Looks like the PDNS API expects same-type records to be updated as a single 'rrset' in the JSON request (using multiple 'records') instead of over multiple 'rrsets'.  This issue will prevent the creation of any records (after the first one) that share a name type, this includes NS and MX entries as well as multiple A and AAAA entries used for round-robin.

This is what the PDNS API expects:
```javascript
{
  "rrsets": [
    {
      "name": "test.com",
      "type": "MX",
      "changetype": "REPLACE",
      "records": [
        {
          "content": "10 mx1.test.com",
          "disabled": false,
          "name": "test.com",
          "ttl": 3600,
          "type": "MX"
        },
        {
          "content": "20 mx2.test.com",
          "disabled": false,
          "name": "test.com",
          "ttl": 3600,
          "type": "MX"
        }
      ]
    }
  ]
}
```

And this is what the PowerDNS-Admin code sends:

```javascript
{
  "rrsets": [
    {
      "name": "test.com",
      "type": "MX",
      "changetype": "REPLACE",
      "records": [
        {
          "content": "10 mx1.test.com",
          "disabled": false,
          "name": "test.com",
          "ttl": 3600,
          "type": "MX"
        }
      ]
    },
	{
	  "name": "test.com",
      "type": "MX",
      "changetype": "REPLACE",
      "records": [
        {
          "content": "20 mx2.test.com",
          "disabled": false,
          "name": "test.com",
          "ttl": 3600,
          "type": "MX"
        }
      ]
	}
  ]
}
```

The code needing the fix is in 'app/models.py'.

ivanfilippov:
I'm not a Python wizard by any means, but I've created a pull request with my solution to this problem. Anyone else wanting to fix the issue can make the code modifications listed in the pull request (https://github.com/ngoduykhanh/PowerDNS-Admin/pull/7)

ngoduykhanh:
Thanks @mortzu for raising this issue and @ivanfilippov for the pull request. I've tried the code in your pull request, it seems not working in all case yet. There was a case that I lose all duplicate records.

I also pushed a new change for fixing this bug, see commit c94e24c98574da2595729547c9e41decce7ffa26

ivanfilippov:
[...]

PR #7, opened 2016-02-11
Fix issue #5. Unable to enter multiple records with same type.

ivanfilippov:
This allows multiple records of the same name and type to be added via the PowerDNS-Admin WebUI. This has been tested successfully on my set up.


ivanfilippov:
Closing, fixed in master see discussion in issue #5

PR #14, opened 2016-03-14
Add code to generate options in 'table-editable.js' to match RECORDS_…

ivanfilippov:
…ALLOW_EDIT in the config.py file.

This change removes the need for a user to manually edit the template files when adding or removing allowed-to-edit record types. Please review and comment.


ngoduykhanh:
Your change looks good to me. Thank you for your contributing.

Issue #28, opened 2016-04-19
Break lines when displaying records, which are very long


piwats:
When you have a long record, f.e. a dkim key which can go up to 254 chars, PowerDNS-Admin does not break the lines in order to be more compact.

This affects how all the records are displayed, that is that the 'Data' field becomes unnecessary long.

ivanfilippov:
This is a good idea.  I have a sample implementation here https://github.com/ivanfilippov/PowerDNS-Admin/tree/width_fix

The end result looks like this:
![wrapping](http://i.imgur.com/e3kAcxD.jpg)

@ngoduykhanh Please review to let me know if my implementation works for you and I'll submit a PR.

ngoduykhanh:
@ivanfilippov : for long line data, after pressing `edit` button i got no data. I think it is because the data field is currently a input text filed, not a text area.

I tried with:
Chrome: 49.0.2623.112
Safari: 9.1

ivanfilippov:
Weird, I'm not seeing that, I'm testing on:

Chrome: 50.0.2661.75
Firefox: 32.0.1
EDGE: 25.10586.0.0

I think it shouldn't affect the input field because they have static widths but I'm going to look into it.

ngoduykhanh:
Yes, it's weird, it tried on Firefox 45.0.2 too. All my test is on MacOS. I will try on a windows machine also.

Sample data:
```
v=DKIM1; p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDfl0chtL4siFYCrSPxw43fqc4zOo3N+Il220oK2Cp+NZw9Kuvg8iu2Ua3zfbUnZWvWK4aEeooliRd7SXIhKpXkgkwnAB3DGAQ6+/7UVXf9xOeupr1DqtNwKt/NngC7ZIZyNRPx1HWKleP13UXCD8macUEbbcBhthrnETKoCg8wOwIDAQAB
```

ivanfilippov:
I've replicated it, has something to do with the doublequotes around that data (I was testing with long string without quotes).

ivanfilippov:
It happens in master as well, not sure when that broke, because I've created and modified TXT and SPF records in my environment before and all mine still have the double-quotes.

So the issue is that the double quotes get parsed as part of the HTML because the first double-quote closes the `value="` tag like so:
[...]

PR #33, opened 2016-04-21
Fix blank Data field when data has double-quotes in it.

ivanfilippov:
This should fix the blank input field discussed in unrelated issue #28. This was the quickest place to fix it I could find and tested ok in my prod environment with a few TXT and SPF records.

@ngoduykhanh please review and test.


app/static/admin/pages/scripts/table-editable.js
```diff
@@ -32,7 +32,7 @@ var TableEditable = function () {
             jqTds[1].innerHTML = '<select class="form-control" id="record_type" name="record_type" value="' + aData[1]  + '"' + '>' + record_types + '</select>';
             jqTds[2].innerHTML = '<select class="form-control" id="record_status" name="record_status" value="' + aData[2]  + '"' + '><option value="false">Active</option><option value="true">Disabled</option></select>';
             jqTds[3].innerHTML = '<select class="form-control" id="record_ttl" name="record_ttl" value="' + aData[3]  + '"' + '><option value="60">1 minute</option><option value="300">5 minutes</option><option value="1800">30 minutes</option><option value="3600">60 minutes</option><option value="86400">24 hours</option></select>';
-            jqTds[4].innerHTML = '<input type="text" class="form-control input-small advance-data" value="' + aData[4] + '">';
+            jqTds[4].innerHTML = '<input type="text" class="form-control input-small advance-data" value="' + aData[4].replace(/\"/g,"&quot;") + '">';
```
ngoduykhanh: What do you think if we also add `style="display:table-cell; width:100% !important"` for this field? It will make the text field width fit the current column for better editing.

ivanfilippov: That's a good idea, do you want me to append it to this PR or make a new one?

ngoduykhanh: You can append it into this PR directly.

PR #49, opened 2016-06-15
Optimize domain sync

CaptainQwark:
Hi,

Thanks for creating pdns-admin!

We have 1300+ domains and the sync that is preformed each time you view the domain list was terribly slow for us. The main problem was that each domain was updated each time, regardless whether there were actual changes or not. Besides that, each domain was queried from the db, but the entire list was already loaded at line 427.

So I reworked the domain list to a dict in line 429, and made the sql updates dependent on actual changes / additions.

Please consider these changes for inclusion, as I think it may help others too. 

Thanks,
Alex.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -426,6 +426,7 @@ class Domain(db.Model):
         """
         db_domain = Domain.query.all()
         list_db_domain = [d.name for d in db_domain]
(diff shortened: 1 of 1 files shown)


ngoduykhanh:
Hi @CaptainQwark ,

Your changes look good to me. Merging the PR.... Thank you for your contribution.

PR #60, opened 2016-07-01
fix for pdns 4.0.0-rc2+ apply changes problems

CaptainQwark:


app/lib/utils.py
```diff
@@ -64,6 +64,9 @@ def fetch_json(remote_url, method='GET', data=None, params=None, headers=None):
     if method == "DELETE":
         return True
 
+    if r.status_code == 204:
```
ngoduykhanh: Why should we put the `status_code` checking here? I think it better to be in `fetch_remote` function, no?

CaptainQwark: You're right, that is a better place to do this. But looking at that code I'm puzzled why my solution even works. It should raise an error for 204 replies. 

(Looks-up stuff) OK. It doesn't because the `requests.raise_for_status()` only raises an error of there actually was one (4xx, 5xx) and 204 isn't an error. 

That bit of code in `fetch_remote` looks a bit weird to me. I guess you do not raise on 400 and 422 because the error info in the returned json provides additional info?
Perhaps its better not to raise on any 4xx / 5xx reply that has and 'error' key in the reply. That way you don't have to worry about the specific status_code numbers (that might be different for different versions of the pdns api)

Or do all the error handling in one place and get the additional info for the error item in `fetch_remote` or `fetch_json` and raise with the additional info, so that the callers in models.py do not have to repeat this error handling. 

Or, while thinking about this, probably the best solution; maybe make `fetch_json` actually **always** return json including an (fabricated) error status, and not relying on actually raising and catching errors in code as they hide actual errors, which make debugging very hard. See [https://realpython.com/blog/python/the-most-diabolical-python-antipattern/](url)

Issue #62, opened 2016-07-01
Dyndns - /nic/checkip.html doesn't seem to work


trenb:
All I get is "None" as a response when I call it. Am I missing something? I thought this was supposed to return the IP you're coming from.
```
mine:~ tren$ curl -v "http://10.21.4.1:9393/nic/checkip.html"
*   Trying 10.21.4.1...
* Connected to 10.21.4.1 (10.21.4.1) port 9393 (#0)
> GET /nic/checkip.html HTTP/1.1
> Host: 10.21.4.1:9393
> User-Agent: curl/7.48.0
> Accept: */*
>
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Content-Type: text/html; charset=utf-8
< Content-Length: 4
< Server: Werkzeug/0.11.10 Python/2.7.5
< Date: Fri, 01 Jul 2016 21:17:19 GMT
<
* Closing connection 0
None
```

Thank you!

trenb:
#63 Fixes this for me
```
mine:~ tren$ curl -v "http://10.21.4.1:9393/nic/checkip.html"
*   Trying 10.21.4.1...
* Connected to 10.21.4.1 (10.21.4.1) port 9393 (#0)
> GET /nic/checkip.html HTTP/1.1
> Host: 10.21.4.1:9393
> User-Agent: curl/7.48.0
> Accept: */*
>
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Content-Type: text/html; charset=utf-8
< Content-Length: 10
< Server: Werkzeug/0.11.10 Python/2.7.5
< Date: Fri, 01 Jul 2016 22:04:29 GMT
<
* Closing connection 0
10.21.4.16
```

trenb:
lol, you're too fast for me! Thanks for looking into this and fixing it.

ivanfilippov:
NP, I'll wait for @ngoduykhanh to accept the PR.

PR #63, opened 2016-07-01
Regarding Issue #62 - Fix to provide client IP

trenb:
use request.environ.get instead of request.headers.get


ivanfilippov:
Hi @trenb 

Thanks for the PR :)

This would break showing the IP when PDNS-Admin is behind a reverse proxy. See PR #64

trenb:
Ah, cool. I'll close this one to PR #64

PR #65, opened 2016-07-01
Replace direct 'click' event binding with delegated binding. Fixes #58.

### LINKED ISSUE #58: Info buttons on items in history do not work on pages other than the first
If I click "info" on any history items on page 1, it displays as expected. But once on page 2, the info button does nothing. 

I am on git commit 17be7ba, running on CentOS 7. Please let me know if you need more info or details.

Thanks!

ivanfilippov:
Direct binding only works for elements already in the DOM, delegated
binding works for all elements that match a filter even if created after
the DOM is fully loaded.

This fixes all click events that have may be created dynamically (when DataTables change pages or sort settings, etc)

### DIFF
--- a/app/templates/admin_history.html
+++ b/app/templates/admin_history.html
@@ -71,8 +71,7 @@
         "info" : true,
         "autoWidth" : false
     });
-
-    $(".history-info-button").click(function() {
+    $(document.body).on('click', '.history-info-button', function() {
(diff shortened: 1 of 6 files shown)


ngoduykhanh:
Merged changes. Thanks @ivanfilippov  for fixing this issue.

PR #71, opened 2016-07-03
Fix width of DNSSEC buttons. Make Enabled button green. Fixes #68.

### LINKED ISSUE #68: Enable/disable DNSSEC
Hi,

Slightly relevant to #2 -- is PowerDNS-Admin supposed to automatically detect if a zone is DNSSEC-enabled?

Currently I have a zone that I have DNSSEC-enabled (via the proper PowerDNS utilities), and the different keys are present. The domain is still listed as "Disabled" within PowerDNS-Admin. However, clicking the "Disabled"-button, it shows the relevant DNSSEC-information.

I looked at the code, and I can't find anywhere where the ``` dnssec``` option would be set to ```true```. The only place I could find, was in ```Domain.update()```, but since there are no place to set true/false within the GUI, this option is never set (regardless if ```Domain.get_domain_dnssec()``` is called when clicking on the "Disabled" button).

ivanfilippov:


ngoduykhanh:
Merged thanks @ivanfilippov .

Issue #74, opened 2016-07-05
TTL for NS-records is "stuck"


jallakim:
Hi,

Trying to change the NS-records for my domains. They are defaulted to 3600 (1 hour). Trying to change these to 86400 (24 hours). I set the value (for all the NS-records), hit "Save", then "Apply". Update successful. Refresh domain/page, old values are displayed (and is also what resides in the DB).

PowerDNS 4.0.0rc2 + PostgreSQL DB backend.

jallakim:
1. "Apply"-button calls for ```/domain/$domain/apply```, which hits ```record_apply()```.
2. Records sent to ```Record.apply()``` has new TTL value (86400).
3. Records in ```for r in new_records:``` have new TTL value.
4. ```final_records``` does **not** have new TTL value (it's now been replaced with the old value).

jallakim:
Culprit seems to be related to ```NEW_SCHEMA```, and in turn changes done in #5.

jallakim:
```
if NEW_SCHEMA:
            records = sorted(records, key = lambda item: (item["name"], item["type"]))
            for key, group in itertools.groupby(records, lambda item: (item["name"], item["type"])):
                new_record = {
                        "name": key[0],
                        "type": key[1],
                        "ttl": records[0]['ttl'],
                        "changetype": "REPLACE",
                        "records": []
                    }
```

This sets the TTL of a record to whatever is stored first in ```records``` (after the sort). Probably not what we want...

jallakim:
I've patched it successfully, but there needs to be more logic in ```Record.apply()```. The API in 4.0.0 seems to limit the TTL to be set per RRset-entry, which means that the logic should handle each TTL for each type of RRset.

https://doc.powerdns.com/md/httpapi/api_spec/#url-apiv1serversserver95idzoneszone95id

E.g. if we have one NS-record with TTL 3600 and another NS-record with TTL 86400, they should be put into two RRset's.

jallakim:
How is the API supposed to handle different TTL's?

According to https://doc.powerdns.com/md/httpapi/api_spec/#url-apiv1serversserver95idzoneszone95id ;

[...]

PR #95, opened 2016-07-26
Use correct TTL in updates

jallakim:
Currently the same TTL is used for all records in the RRset. This *could* cause entries to get deleted whenever certain circumstances are present (e.g. if you were to set `@ IN NS ns1.foo.bar` with `TTL=3600` and `@ IN NS ns2.foo.bar` with `TTL=1800`).

This change sets the TTL to be the same for all records with the same name and type.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -858,8 +858,8 @@ class Record(object):
                 record = {
                             "name": r['name'] + '.',
                             "type": r['type'],
-                            "ttl": r['ttl'],
                             "changetype": "REPLACE",
+                            "ttl": r['ttl'],
                             "records": [
                                 {
                                     "content": r['content'],
@@ -888,14 +888,14 @@ class Record(object):
 
         # Adjustment to add multiple records which described in https://github.com/ngoduykhanh/PowerDNS-Admin/issues/5#issuecomment-181637576
         final_records = []
-        records = sorted(records, key = lambda item: (item["name"], item["type"]))
-        for key, group in itertools.groupby(records, lambda item: (item["name"], item["type"], item["changetype"], item["ttl"])):
+        records = sorted(records, key = lambda item: (item["name"], item["type"], item["changetype"]))
+        for key, group in itertools.groupby(records, lambda item: (item["name"], item["type"], item["changetype"])):
             if NEW_SCHEMA:
                 new_record = {
                         "name": key[0],
                         "type": key[1],
                         "changetype": key[2],
-                        "ttl": key[3],
+                        "ttl": None,
                         "records": []
                     }
                 for item in group:
@@ -905,6 +905,8 @@ class Record(object):
                         if temp_content.strip()[-1:] != '.':
                             temp_content += '.'
 
+                    if new_record['ttl'] is None:
+                        new_record['ttl'] = item['ttl']
                     new_record['records'].append({
                         "content": temp_content,
                         "disabled": temp_disabled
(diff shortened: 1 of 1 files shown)


jallakim:
Replaces #75.

jallakim:
Fixes #74.

app/models.py
```diff
@@ -888,14 +888,14 @@ def apply(self, domain, post_records):
...
-        if NEW_SCHEMA:
-            records = sorted(records, key = lambda item: (item["name"], item["type"]))
-            for key, group in itertools.groupby(records, lambda item: (item["name"], item["type"])):
+        records = sorted(records, key = lambda item: (item["name"], item["type"]))
+        for key, group in itertools.groupby(records, lambda item: (item["name"], item["type"], item["changetype"], item["ttl"])):
+            if NEW_SCHEMA:
                 new_record = {
                         "name": key[0],
                         "type": key[1],
-                        "ttl": records[0]['ttl'],
```
jallakim: This sets the TTL to the same for all entries in the RRset, using the first TTL-value in the records-list.

app/models.py
```diff
@@ -888,14 +888,14 @@ def apply(self, domain, post_records):
 
         # Adjustment to add multiple records which described in https://github.com/ngoduykhanh/PowerDNS-Admin/issues/5#issuecomment-181637576
         final_records = []
-        if NEW_SCHEMA:
-            records = sorted(records, key = lambda item: (item["name"], item["type"]))
-            for key, group in itertools.groupby(records, lambda item: (item["name"], item["type"])):
+        records = sorted(records, key = lambda item: (item["name"], item["type"]))
+        for key, group in itertools.groupby(records, lambda item: (item["name"], item["type"], item["changetype"], item["ttl"])):
```
ngoduykhanh: `item["ttl"]` shouldn't be in the `groupby`, otherwise this will split into 2 array (in case 2 records have same Name-Type but difference TTL) then one of them will be deleted. 

Tested on pdns 4.0.0~alpha2-3build1

jallakim: Yes, you are correct. I actually had the same issues (one of them would be deleted), but upon more tests, it didn't happen. I think I might've tested that last part on the wrong branch.

app/models.py
```diff
@@ -888,14 +888,14 @@ def apply(self, domain, post_records):
 
         # Adjustment to add multiple records which described in https://github.com/ngoduykhanh/PowerDNS-Admin/issues/5#issuecomment-181637576
         final_records = []
-        if NEW_SCHEMA:
-            records = sorted(records, key = lambda item: (item["name"], item["type"]))
-            for key, group in itertools.groupby(records, lambda item: (item["name"], item["type"])):
+        records = sorted(records, key = lambda item: (item["name"], item["type"], item["changetype"]))
+        for key, group in itertools.groupby(records, lambda item: (item["name"], item["type"], item["changetype"])):
```
jallakim: Only grouping by name+type (and changetype, but that shouldn't really matter since we set it statically in the previous part -- just felt it was The Right Way™ to not set it statically when we already did that).

app/models.py
```diff
@@ -905,18 +905,20 @@ def apply(self, domain, post_records):
                         if temp_content.strip()[-1:] != '.':
                             temp_content += '.'
 
+                    if new_record['ttl'] is None:
+                        new_record['ttl'] = item['ttl']
```
jallakim: Set TTL to the first TTL of that name+type tuple (which is the record sorted topmost in the GUI).

ngoduykhanh:
Looks good to me. Merging.
Thanks @jallakim

PR #97, opened 2016-08-02
Fix route when the zone name containing slash character.

dorion:
We delagate many sub zones and foward zones so in our database containing slash charater in the revers zone names. Whitout this fix I give back 404 error page.


ngoduykhanh:
@dorion : Is that slash used for netmask portion of zone names?

dorion:
Yes of course, like here https://www.ietf.org/rfc/rfc2317.txt.
Because I made this fix!

ngoduykhanh:
ok, merging. Thanks for your fixing.

dorion:
Thanks your merging :-)

PR #102, opened 2016-08-11
Add auto PTR support for the A and AAAA records.

dorion:
Use the powerDNS minial one-way auto-ptr functionality.

Feel free to modify if it isn't harmony with the flask philosopy.

Usage: when add or modify an A or AAAA record appear an modal windonw to help to set auto PTR.


ngoduykhanh:
@dorion : Thank your for your contribution. I tried your branch and it works. But seems we stil need more improvement:
- How do people know if a record has PTR (from Record management page). The check box wasn't checked after PRT created.
- In my opinion, appending the string "Auto PTR!" into record's value doesn't look nice. Should it be a checked box in the same row instead ?
- There a several `console.log` in your code, I think you used it for debugging. Can we remove it ?

jallakim:
Another thing; should the auto-ptr be taken into consideration when deleting forward records? (i.e. deleting a forward record that previously generated a PTR, should the PTR also be deleted?). Not really sure how pdns handles that, or if PDA should handle it?

dorion:
If you check I use the PowerDNS built-in auto PTR feature, which is poor documented and very simple implementations of the functionality. I only use "set-ptr" when create or update a RR. But the API won't give back any information about the history of the auto PTR. I thinking the auto PTR work only on-the-fly and overwrite the before PTR record whitout any notification and the server is not track this event!

I will work on your another two suggestions, but now I am on holiday :-)

jallakim:
This pull-request should be cleaned up/rebased/whatever. I guess it's still a work-in-progress, but..

petersipos:
Hello @dorion I've developed a new solution to this topic based on new aproach. You can see on the #156 pull request.

ngoduykhanh:
This feature was implemented by @petersipos . I am closing this PR. Thanks.

PR #103, opened 2016-08-12
support github oauth2 login

timfeirg:
this PR introduces github oauth2 login by @cmgs, I did some minor improvements on code style.

Please tell us if you're happy with the functionality, or if this PR needs more work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ logfile.log
 db_repository/*
 upload/avatar/*
 tmp/*
+.ropeproject
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,7 +1,7 @@
 from werkzeug.contrib.fixers import ProxyFix
-from flask import Flask
-from flask.ext.login import LoginManager
-from flask.ext.sqlalchemy import SQLAlchemy
+from flask import Flask, request, session, redirect, url_for
+from flask_login import LoginManager
+from flask_sqlalchemy import SQLAlchemy
 
 app = Flask(__name__)
 app.config.from_object('config')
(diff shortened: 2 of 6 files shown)


ngoduykhanh:
Thank you @timfeirg  and @CMGS  for your contribution. I have few comment:
- Why do you change from `#!flask/bin/python` to `#!/usr/bin/python` while we are using virtualenv ? 
- In case that Github Oauth is enabled, I see that the login form is changed. Text "Github oauth login" and "Create an account" are in the same line. Can we make it look better ?
- Can you put commit message in English instead of Chinese so everybody can understand easily.
- DNS is an important system and usually used internally. I don't know if it is a good idea to have OaAuth for it. If so, why only Github?.

CMGS:
let me explain it.

I changed headline macros because I think we should consider scripts are using in normal case. not everyone used virtualenv in general way, for example I used virtualenvwrapper instead. however this change was not important in this PR, if u dislike it I can revert to your version.

in our team, we used github private plan for many years, this year we bought github enterprise instead of it. if we want to login internal systems with a single account, using it's account will be more convenient for us than using LDAP (there is no 'SA' in our team). as I know, many small/medium team like us, that's why we implement github oauth login not a normal oauth login.

at last I will change login form ASAP, thank for your replying.

ngoduykhanh:
> I changed headline macros because I think we should consider scripts are using in normal case. not everyone used virtualenv in general way, for example I used virtualenvwrapper instead. however this change was not important in this PR, if u dislike it I can revert to your version.

Yes, I agree with you. But this is not the main purpose of this PR so please revert it. We can change it later with updated README/WIKI also.


> in our team, we used github private plan for many years, this year we bought github enterprise instead of it. if we want to login internal systems with a single account, using it's account will be more convenient for us than using LDAP (there is no 'SA' in our team). as I know, many small/medium team like us, that's why we implement github oauth login not a normal oauth login.

Understood. As far as it is optional and can be configured with config file. I am ok with this feature.


> at last I will change login form ASAP, thank for your replying.

Thanks 👍

timfeirg:
I've squashed the commits, revert the shebang lines, translated the commit message, and add a `<br>` between the two `<a>` tags, I knew too little about writing webpages.

@ngoduykhanh let me know if anything still needs change

jallakim:
Can we have `GITHUB_OAUTH_ENABLE = False` as default in `config_template.py` since the majority won't use it.

timfeirg:
done @jallakim 
>Can we have GITHUB_OAUTH_ENABLE = False as default in config_template.py since the majority won't use it? (now it's set to True).

ngoduykhanh:
Looks good to me. Merging. Thanks @timfeirg .

PR #104, opened 2016-08-12
Users could use SQLite instead of MySQL

wzyboy:
PowerDNS-Admin uses SQLAlchemy as ORM, which supports various database backends. `MySQL-python` should not be listed as a must-have depenedency in `requirements.txt` since it needs compiling and complex setup. For testing purpose, users could use SQLite, which is a built-in function in most Python distributions.

This PR removes `MySQL-python` dependency and instructs users to try SQLite if they do not want to install MySQL / MySQL-python.

My tests show that PowerDNS-Admin works perfectly with SQLite backend.


ngoduykhanh:
@wzyboy : I agree with you. Merged your PR. Thanks.

Issue #98, opened 2016-08-04
Delete domain does not work


michielvisser:
I just installed PowerDNS admin and I like the tool. However, I cannot delete any domains I've created, the same as in issue #89. My setup is based on CentOS 7, powerDNS 3.48 and MySQL as the backend. SELinux has been disabled.

I can delete all the records from a domain, that works fine. However, going to the admin button for the domain I want to delete, the click delete and confirm, nothing happens. The dashboard just gets reloaded and the doman is still there.

For now I've deleted the domain manually from both PowerDNS and PowerDNS-admin database.

vauxia:
The log says, "IOError: [Errno 32] Broken pipe" when this happens.

ngoduykhanh:
I couldn't reproduce @michielvisser 's issue with the same setup. It works fine for me. Does the history log show anything?

@vauxia : Try to restart application and delete the domain again. It looks like your application hangs. You can also set `threaded=True` in the `app.run()` function in `run.py` file to improve the application performance.

michielvisser:
Tried today with the latest commit, but this does not change the issue. The history log does not show that much, here is the output. Actions I did:
- Logged in with my user
- Clicked on 'admin' for the domain I wanted to delete
- Clicked 'Delete domain [name]'
- Confirmed the delete

Then I got the dashboard back and nothing has happened. I also cannot see a delete request of some kind going to the PDNS server.

Here is the log output I have of the above:

(flask) [root@powerdns PowerDNS-Admin]# ./run.py
[INFO]  * Running on http://192.168.50.25:80/ (Press CTRL+C to quit)
[INFO]  * Restarting with stat
[WARNING]  * Debugger is active!
[INFO]  * Debugger pin code: 103-133-745
[INFO] 192.168.50.60 - - [16/Aug/2016 07:35:24] "GET / HTTP/1.1" 200 -
[INFO] 192.168.50.60 - - [16/Aug/2016 07:35:24] "GET /static/adminlte2/bootstrap/css/bootstrap.min.css HTTP/1.1" 200 -
[INFO] 192.168.50.60 - - [16/Aug/2016 07:35:24] "GET /static/adminlte2/dist/css/AdminLTE.min.css HTTP/1.1" 200 -
[INFO] 192.168.50.60 - - [16/Aug/2016 07:35:24] "GET /static/adminlte2/plugins/iCheck/square/blue.css HTTP/1.1" 200 -
[...]

PR #106, opened 2016-08-15
Fix domain delete action if uwsgi path mount is not /

vdchuyen:
For ex: mount = /pdsn-admin=run.py


ngoduykhanh:
Nice catch @vdchuyen . This may fix the issue #98 . Merging. Thanks.

PR #111, opened 2016-08-16
Clean up SQL and AD config

elitest:
This PR suggests that the config file should have an example for AD as well as LDAP.  It also suggests that the SQL config should more closely mirror what is suggested for MySQL config in the README(e.g. user=powerdnsadmin instead of user=root).  It also splits the config into multiple variables.  Let me know what you think.

Cheers


jallakim:
Since you edited both those files; maybe we can just remove the `config_template_docker.py` completely? The only (major) difference is that the Docker-version has `LOG_FILE = '/dev/stdout'`, which, imho, can just be a line that is commented out like;

```
LOG_FILE = 'logfile.log'
# For Docker, set the following
#LOG_FILE = '/dev/stdout'
```

... or similar. The fact that `config_template_docker.py` is completely outdated (compared to `config_template.py`) suggests that it might be a good idea to remove it (-:

elitest:
Done.

ngoduykhanh:
Looks good to me. Merging. Thanks @elitest.

PR #113, opened 2016-08-18
Allow to change root domain record via dyndns

ssendev:
It was not possible to update `example.com`. only `foo.example.com`

### DIFF
--- a/app/views.py
+++ b/app/views.py
@@ -762,12 +762,12 @@ def dyndns_update():
     domain = None
     domain_segments = hostname.split('.')
     for index in range(len(domain_segments)):
-        domain_segments.pop(0)
         full_domain = '.'.join(domain_segments)
         potential_domain = Domain.query.filter(Domain.name == full_domain).first()
         if potential_domain in domains:
             domain = potential_domain
             break
+        domain_segments.pop(0)
 
     if not domain:
         history = History(msg="DynDNS update: attempted update of %s but it does not exist for this user" % hostname, created_by=current_user.username)


ngoduykhanh:
Thanks for catching this issue. You also need to handle the `example.com` with `SOA` type unless `example.com` (A record) won't be updated.

Quic fix [here](https://github.com/ngoduykhanh/PowerDNS-Admin/commit/7bcdc3cfd53c7e1ff186fca5775ea5289ea34d20) in another branch. cc @ivanfilippov

ssendev:
Should i add your changes or do you just merge your branch?

ngoduykhanh:
If my changes is ok, i will merge it by myself. Just wanna hear from @ivanfilippov since he suggested this feature and may using it.

ivanfilippov:
@ngoduykhanh Looks good to me, this is a great addition.

Sorry for the delay in response and not being to help out @ngoduykhanh, work has been extra busy.

ssendev:
do your changes prevent updating of AAAA records?

ngoduykhanh:
> do your changes prevent updating of AAAA records?

Right. only IPv4 is supported. Do we have dynamic ipv6 in real use cases ?

ssendev:
I know some people who have ipv6 at home and my router supports sending ipv6 addresses. In fact it allows sending of v4 and v6 at the same time but I am not sure there are isps who give two ips to their customers.

joniw:
Hello,

what is the state of this PR? 

My ISP runs a dualstack environment, i get an ipv4 and an ipv6 address.

Issue #115, opened 2016-08-19
Reset search field on "add record" event


adrpp:
Hello,

not really a bug, but might confuse a bit. Inside of domain where existent RR's are shown, when entering a search string into the search field and afterwards clicking on "add record" shows 'nothing'.

This is by design, as the newly created (empty) RR field doesn't match the search string.
Maybe want to reset search field after clicking on "add record" button to workaround this behavior ?

Rgds,
Adrian

jallakim:
Hi,

This is related to #94. Your suggestion would fix the issue I described (if you have an active search), but would not fix the problem if you've altered the sorting of the rows (as OP of #94 describes).

Both these issues should probably be addressed in the same fix, hence, I'm not sure clearing the search field on "Add record" would be enough...

ngoduykhanh:
This issue was fixed by @jallakim . Please get the latest code. Thanks.

PR #117, opened 2016-08-19
Fix issues when adding a record, and sort and/or search is involved

### LINKED ISSUE #94: Issue with add record field when sort order changed
When adding a record to a domain, if you change the sort order "highest to lowest", the entry dialogue is moved to the bottom of the list. Changing the sort order should not change where the entry dialogue is located.

Let me know if you need any more information.

jallakim:
Previously this caused the newly added record to either appear at the bottom, or not appear at all. Now it will always be added at the top, and whatever search present is cleared.

Fixes #94 and #115.

### DIFF
--- a/app/static/custom/js/custom.js
+++ b/app/static/custom/js/custom.js
@@ -90,7 +90,7 @@ function editRow(oTable, nRow) {
        var record_type = records_allow_edit[i];
        record_types += "<option value=\"" + record_type + "\">" + record_type + "</option>";
     }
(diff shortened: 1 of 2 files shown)


jallakim:
Also cleaned up some of the other things (missing modal with `TODO:`, and the `<td>`-entries).

jallakim:
Also fixed so that when adding a new record, the first text-field is focused (minimizes mouse/tab-interaction (-: ).

jallakim:
@ngoduykhanh *notify*

ngoduykhanh:
Looks good to me. Thanks @jallakim for this adjustment. Merging...

PR #118, opened 2016-08-19
Pretty IPv6 PTR. Use the actual IPv6 address when editing PTR.

jallakim:
Rather than dealing with ip6.arpa-dotted-strings from hell, you can
now edit IPv6 PTR-records using the IPv6 address.


ngoduykhanh:
Looks good to me. Merging.  Thanks @jallakim .

ngoduykhanh:
I also did a small change to not require `PRETTY_IPV6_PTR` by default. Commit 3c31fabf96dddc5bd23cbc19994199340e70a261

PR #121, opened 2016-08-21
Pretty history details

### LINKED ISSUE #99: History information detail unreadable on large changes
I just added a bunch of records to a domain. When I view the change history details for that edit, it is completely unreadable for humans:

image

System is based on CentOS 7, PowerDNS 3.4.8, latest commit of PowerDNS-admin.

jallakim:
Fixes #99.

### DIFF
--- a/app/static/custom/css/custom.css
+++ b/app/static/custom/css/custom.css
@@ -14,3 +14,21 @@ table td {
 .btn {
     padding: 5px 4px !important;
 }
+
+/* Pretty JSON */
+.json-pre {
+   background-color: ghostwhite;
+   border: 1px solid silver;
+   padding: 10px 20px;
+   margin: 20px;
+}
+.json-key {
+   color: brown;
+}
+.json-value {
+   color: navy;
+}
+.json-string {
+   color: olive;
+}
+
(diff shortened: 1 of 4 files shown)


jallakim:
The files `custom.css` and `custom.js` had some weird file type or something, so that new lines added to the file got `^M` as carriage return (in stead of '\r' or whatever). I did a `dos2linux` on them, hence, they seem to have been "completely replaced". Only the last few lines were altered, tho.

ngoduykhanh:
Thanks @jallakim. It looks much better with pretty json display. Can we also use the scroll inside the modal incase the list is long?

jallakim:
The scroll outside the modal works pretty much as one inside the modal? At least on my computer in Chrome...

jallakim:
@ngoduykhanh, doesn't seem like it's a good way to set scroll within the modal (modal body). Only way I've found is to set a static height for the modal-box, but that's not really something we want since people have different sized screens. Using % as height is not possible because it's defined by the parent, which, in this case, also contains the content.

ngoduykhanh:
@jallakim : Ok, so let stick with the your design. We can do more adjustment in the future. Thank you very much for your contribution.

PR #124, opened 2016-08-23
set user object id after created

timfeirg:
this fix a bug in `create_local_user`, which breaks github login.

the commit is line 271, the rest is whitespace stripping.


ngoduykhanh:
Merging. Thanks @timfeirg for catching that.

timfeirg:
credit goes to @CMGS

PR #125, opened 2016-08-23
models refactoring

timfeirg:
authored by @CMGS , this commit intends to make `models.py` more pythonic

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -172,73 +172,68 @@ class User(db.Model):
                 if user_info.password and self.check_password(user_info.password):
                     logging.info('User "%s" logged in successfully' % self.username)
                     return True
-                else:
-                    logging.error('User "%s" input a wrong password' % self.username)
-                    return False
-            else:
-                logging.warning('User "%s" does not exist' % self.username)
+                logging.error('User "%s" input a wrong password' % self.username)
                 return False
 
-        elif method == 'LDAP':
+            logging.warning('User "%s" does not exist' % self.username)
+            return False
+
+        if method == 'LDAP':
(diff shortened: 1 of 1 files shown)


ngoduykhanh:
Thanks @CMGS and @timfeirg for your work. It looks like we have problem with user creation step in this refactoring. Password hashed isn't created after user sign up.

It was handled by this [original code](https://github.com/ngoduykhanh/PowerDNS-Admin/blob/master/app/models.py#L268)

ngoduykhanh:
Also, have you tested with LDAP authentication? I am not sure if it won't break anything if we don't assgin user's id after create them [at this line](https://github.com/ngoduykhanh/PowerDNS-Admin/pull/125/files#diff-413e9aa8961126d20e2d90befa13e0f1L245)

timfeirg:
fixed the password issue at [this line](https://github.com/ngoduykhanh/PowerDNS-Admin/pull/125/files#diff-413e9aa8961126d20e2d90befa13e0f1R265)

>Thanks @CMGS and @timfeirg for your work. It looks like we have problem with user creation step in this refactoring. Password hash isn't created after user sign up.

as for `self.id`, I don't think it's a problem, `self.id` is set during `__init__`, so when executing `self.create_user`, all the user attributes gets committed to the database. @CMGS had tried this out and confirmed this behavior.

>Also, have you tested with LDAP authentication? I am not sure if it won't break anything if we don't assgin user's id after create them at this line

ngoduykhanh:
Looks  good to me now. Merging. Thanks.

PR #132, opened 2016-08-31
Log to file only if the file exists

pschiffe:
If LOG_FILE config option contained invalid file, pdns admin wouldn't start.
Also, it wasn't possible to log only to stdout/stderr correctly, because
setting the log file to /dev/stdout would double the logs messages as the pdns
admin is already logging to the stderr.

### DIFF
--- a/app/lib/log.py
+++ b/app/lib/log.py
@@ -27,18 +27,21 @@ class logger(object):
         # set request requests module log level
         logging.getLogger("requests").setLevel(logging.CRITICAL)
 
-        # define handler to log into file
-        file_log_handler = logging.FileHandler(self.logfile)
-        logger.addHandler(file_log_handler)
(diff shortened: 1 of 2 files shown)


jallakim:
Should have more logic;

* if both path/folder and file exists, log as usual
 * if path/folder exists, but not the file; create file
 * if path/folder doesn't exist, exit with error (or at least display error)

User shouldn't have to create the logfile manually (but the path; sure).

pschiffe:
You are right that the file doesn't need to exist prior logger can log to it. I've removed that check. But I wouldn't complicate it with checks for path/folder. My main goal of this PR is let the user skip the log file, because when running in docker or under uwsgi, systemd.. you don't care about log files, only stdout/stderr is important.

ngoduykhanh:
I am ok with this PR. Merging. Thanks @pschiffe .

PR #139, opened 2016-09-12
Fix issue with deleting domain

### LINKED ISSUE #98: Delete domain does not work

Cultti:
Fixes #98 

However it would be nice to get error message if the delete does not work.


ngoduykhanh:
Looks good to me. Thanks @Cultti  for your PR. Merging...

PR #142, opened 2016-09-17
Replace python libraries with better maintained versions

kaechele:
Also fixed encoding issue in password hashing and shebang lines.

Tested and working.


kaechele:
Note: This would reqiure running pip install -r requirements.txt in the virtualenv again

ngoduykhanh:
Looks good to me. Merging. Thank you @kaechele  for your contribution.

PR #143, opened 2016-09-18
Python 3 support

kaechele:
I tested this on Python 2.7 and 3.5 on Fedora 24. Both seem to work as expected performing regular operations.
I would appreciate if someone tested this as well, just to make sure.

@ngoduykhanh Are you okay with removing the explicit version numbers from requirements.txt? During development we should make sure that PowerDNS-Admin always works with the most recent versions. For releases it could make sense to use pip freeze then.


kaechele:
Hmm. For some reason the commit on porting to ldap3 slipped in to the "Make unicode aware" commit :(

ngoduykhanh:
Thanks @kaechele  for your great work. I will help to test these changes also.

about the requirements.txt, I agree that we should keep PowerDNS-Admin always works with the most recent versions but still, we need to keep the number in requirements.txt so everybody can know which version is currently supported.

kaechele:
Okay. Makes sense. Then maybe we can test these changes here and then create a new, clean pull request with all the changes.

ngoduykhanh:
My first test on user creation step was failed. I got the following exception

```
  File "/Users/khanh/code/github/pda-kaechele/app/views.py", line 254, in login
    error = e.message['desc'] if 'desc' in e.message else e
AttributeError: 'DataError' object has no attribute 'message'
```

The exception was raised because the password field because encrypt with length longer than 64  character which we defined in the model - we might need to extend this length limitation.

The exception message also shows that `e` variable doesn't have `message` attribute itself.

ngoduykhanh:
I also tried to extend the password field length in the model. The registration passed. `123` password was encrypted as `\x2432622431322461576f516d56446d2e37376e626f36642e7a73503175396f7875437032596332716b476d4f514553656656334644794f58564e6779` string. But I cannot login. An excption was thrown.

```
  File "/Users/khanh/code/github/pda-kaechele/flask3/lib/python3.5/site-packages/bcrypt/__init__.py", line 86, in hashpw
    raise ValueError("Invalid salt")
ValueError: Invalid salt
```

kaechele:
Strange. I actually tested that. Which platform are you using? Looks like OS X. I will try to reproduce it locally.

ngoduykhanh:
Yes, I am using MacOSX, Python 3.5 installed via virtualenv, Postgresql is DB backend.

kaechele:
I came to the conclusion that before pursuing this further I would like to write some test cases first so we can immediately see if something is not working.
I will follow this guide http://flask.pocoo.org/docs/0.11/testing/

kaechele:
Made some fixes and did more testing on OS X 10.12.1 using Python 3.5.2 from Homebrew.
Everything seems to be working now.

Still didn't get around to writing the test cases yet, since there is only little that we can test without involving a complete test rig with database and pdns servers. So writing an automated testsuite would probably require using Vagrant or something similar to set up the complete pdns / PowerDNS-Admin environment.

benfairless:
Tested your Python3 branch today and received the following message on user registration:

Screenshot of login page

kaechele:
Yup. There is a small error in the calling of the hashlib functions with regard to utf-8 encoding.
This pull request should probably be closed and re-done anyway.

akurach:
any updates?

ngoduykhanh:
Solved in `development` branch. Waiting to merge...

Issue #133, opened 2016-09-02
LDAP password store in db as plain text


unclev0va:
But if I create new user, password stored hashed.
In code:

```
    def create_user(self):
        """
        If user logged in successfully via LDAP in the first time
        We will create a local user (in DB) in order to manage user
        profile such as name, roles,...
        """
        db.session.add(self)
        db.session.commit()
```

I fix this by add  
`self.password = self.get_hashed_password(self.plain_text_password)`
before 
`db.session.add(self)`

jallakim:
The password-logic should probably be removed from LDAP-users completely? Or is the idea that LDAP-users should be able to log in with whatever passord is stored locally if LDAP suddenly becomes unreachable? (provided that they've logged in before and the user was created).

pschiffe:
User's LDAP passwords should never be stored in local DB whatsoever - plain or hashed. That's one of the point of centralized identity management, that even when the app is breached, passwords are still safe.. This issue is a security vulnerability.

jallakim:
@pschiffe, never thought about it from that angle before, but I guess most people utilize LDAP to be able to have the same password "everywhere". I know of several applications that caches users/passwords locally in order to facilitate login in case LDAP is unreachable (if user is deactivated in LDAP, and LDAP is reachable, the users won't be able to log in, of course -- so only in the event of an unreachable LDAP server).

The biggest issue here is that it's stored in plaintext, but I have no issue with removing it completely...

pschiffe:
@jallakim LDAP is used in that way because it's easy to set up and use that way. Caching passwords is definitely a wrong work-around to a problem, which should be solved differently (LDAP replicas). But again, because it's possible to cache it, and relatively easy to do, people do it. But that doesn't mean it's OK, it's definitely insecure.

When you look at the modern standards for "one password everywhere", or single sign on - an oauth or openid connect, you see that in this case, the app never sees the password, so it's not possible to cache it. And that's how it should be.

Please, if it's possible, don't store the LDAP password in pdns-admin db in any way.

ngoduykhanh:
I agree with @pschiffe . The User which created from LDAP or openID should never have a password. LDAP's "local user" in the pdns-admin is only used for role/permission setting. The current code should be changed to remove LDAP user password completely.

PR #146, opened 2016-09-28
Fix issue #133. Store hash as '*' for external users

bergzand:
Set password to '*' for users created by the create_user method. Should cause an invalid password hash for non local users added to the database


petersipos:
+1

ngoduykhanh:
Merging. Thanks @bergzand

PR #150, opened 2016-10-13
re-fix root domain record update via dyndns

mir06:
if hostname is @.domain.tld (provided by the ddns-client) the DNS type has to be A or AAAA, which is checked in the exists method of Record.

Thus CNAME (or other) updates are still possible, and ipv6 records can also be updated.


ngoduykhanh:
Merged. Thanks for your contribution. I will review the `dyndns_adjustment` again then merge to master later.

PR #156, opened 2016-11-16
Feature/automatic reverse domain creation

petersipos:
I've created my own non solution of creating auto PTR records. This way is different from #102 

In my pull request I've solved the automatic reverse domain creation with the user privileges of the original domain. And now I working on the record deleting and updating. In this case it isn't necessary to send set-ptr: auto in PATCH to pdnsd, and it is working on older version <4.0.0 of it. 

I'll create in this pull request the following things:
* [x] safely update, delete the existing ptr records
* [x] ability on gui to choose of creating ptr records from A or AAAA

Please review my solution and discuss.

**UPDATE:**
~~I'll create a NATIVE auto-ptr functionality to my own~~
~~* [ ] Use pdnsd set-ptr record~~

**UPDATE2:**
I see that the native version of auto-ptr functionality isnt working very well in all conditions.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1044,8 +1044,10 @@ class Record(object):
         domain_obj = Domain.query.filter(Domain.name == domain).first()
         domain_auto_ptr = DomainSetting.query.filter(DomainSetting.domain == domain_obj).filter(DomainSetting.setting == 'auto_ptr').first()
         domain_auto_ptr = strtobool(domain_auto_ptr.value) if domain_auto_ptr else False
+
         system_auto_ptr = Setting.query.filter(Setting.name == 'auto_ptr').first()
         system_auto_ptr = strtobool(system_auto_ptr.value)
+        
         if system_auto_ptr or domain_auto_ptr:
             try:
                 d = Domain()
@@ -1053,7 +1055,7 @@ class Record(object):
                     if r['type'] in ['A', 'AAAA']:
                         r_name = r['name'] + '.'
                         r_content = r['content']
-                        temp = re.search('^(([a-f0-9]\.){4}(?P<ipv6name>.+6.arpa)\.?)|(\.(?P<ipv4name>.+r.arpa)\.?)', dns.reversename.from_address(r_content).to_text())
+                        temp = re.search('^((([a-f0-9]\.){4})(?P<ipv6name>.+6.arpa)\.?)|((([0-9]+\.){1})(?P<ipv4name>.+r.arpa)\.?)', dns.reversename.from_address(r_content).to_text())
                         domain_reverse_name = temp.group('ipv6name') if temp.group('ipv6name') != None else temp.group('ipv4name')  
                         d.create_reverse_domain(domain, domain_reverse_name)
                         self.name = dns.reversename.from_address(r_content).to_text().rstrip('.')
@@ -1066,7 +1068,7 @@ class Record(object):
                     if r['type'] in ['A', 'AAAA']:
                         r_name = r['name'] + '.'
                         r_content = r['content']
-                        temp = re.search('^(([a-f0-9]\.){4}(?P<ipv6name>.+6.arpa)\.?)|(\.(?P<ipv4name>.+r.arpa)\.?)', dns.reversename.from_address(r_content).to_text())
+                        temp = re.search('^((([a-f0-9]\.){4})(?P<ipv6name>.+6.arpa)\.?)|((([0-9]+\.){1})(?P<ipv4name>.+r.arpa)\.?)', dns.reversename.from_address(r_content).to_text())
                         domain_reverse_name = temp.group('ipv6name') if temp.group('ipv6name') != None else temp.group('ipv4name')  
                         self.name = dns.reversename.from_address(r_content).to_text()
                         self.type = 'PTR'
--- a/app/models.py
+++ b/app/models.py
@@ -610,7 +610,7 @@ class Domain(db.Model):
         Check the existing reverse lookup domain, 
         if not exists create a new one automatically
         """
-        domain_obj = Domain.query.filter(Domain.name == domain).first()
+        domain_obj = Domain.query.filter(Domain.name == domain_name).first()
         domain_auto_ptr = DomainSetting.query.filter(DomainSetting.domain == domain_obj).filter(DomainSetting.setting == 'auto_ptr').first()
         domain_auto_ptr = strtobool(domain_auto_ptr.value) if domain_auto_ptr else False
         system_auto_ptr = Setting.query.filter(Setting.name == 'auto_ptr').first()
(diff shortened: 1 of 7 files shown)


petersipos:
I finished it, ready to pull ;)

olfway:
@petersipos will it be possible to create one "10.in-addr.arpa" zone instead of "1.0.10.in-addr.arpa", "2.0.10.in-addr.arpa", etc?

petersipos:
@olfway Hmm well, it needs to support A and B type of IP addresses not just C... hmm yeah.
But this feature doesn't support it yet.

olfway:
@petersipos maybe it should do something like this?

- check if "1.0.10.in-addr.arpa" exists and then use it
- check if "0.10.in-addr.arpa" exists and then use it
- check if "10.in-addr.arpa" exists and then use it
if nothing found, then create "1.0.10.in-addr.arpa" and use it

i used 10.0.0.1 as example here, it should do the same for any ip address

jallakim:
Haven't read the code yet -- but does this PR also address IPv6 PTRs?

petersipos:
@jallakim Yes it creates the reverse domain for ipv6 addresses too and also creates the IPv6 PTRs...

@olfway I try to figure out an elegant solution for this request :)

jallakim:
@petersipos, then the discussion you're having (`1.0.10.in-addr.arpa` vs. `10.in-addr.arpa`) is just as relevant (if not *more* relevant) for IPv6 PTRs.

jallakim:
I also have to verifiy if this breaks `PRETTY_IPV6_PTR` (my "problem", not yours, I guess).

petersipos:
@jallakim it works fine with the `PRETTY_IPV6_PTR` settings
Yes I agree, the support of IP classes is relevant, but when we speak about IPv4 adresses we have just 3 class types (A,B,C quiet simple...) but in IPv6 this task began will be more difficult.

The most difficutl question is, how can we give an elegant and user friendly solution in the GUI to manage the required IP classes.

petersipos:
@olfway Yes it will be maybe a simple solution for IPv4 addresses.

olfway:
@petersipos from dns point of view, there is no A, B or C classes, it's just subdomains
so for ipv6 zone "0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa."
you can check exact domain first and if it's not found, strip one subdomain and check again (up to ip6.arpa?)
if nothing found - create "0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa."

petersipos:
yes is it possible, and simple.

petersipos:
@olfway and @jallakim I've finished the corrections what you mentioned. (The same way as you mentioned). Please check it.

Now this modification check the existing reverse lookup domain zones from the highest level, and if find one add the PTR records there. If not exists create the lowest possible reverse lookup domain zone and create the PTRs there. Also works with IPv4 and IPv6.

When I check the IPv6 domain I check just the domains:
1.3.f.7.[b.c.e.f.f.f.9.2.c.0.2.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.e.f.ip6.arpa.] (in thre brackets)
Maybe I may check all of the digits?

olfway:
@petersipos thanks!
from the code, it seems that it will auto-create "in-addr.arpa" zone in case it will not find more specific one
am i right?

petersipos:
no, it whouldn't create in-addr.arpa, just eg. 192.in-addr.arpa 
You need to create an in-addr.arpa zone? Isn't so difficult to modify, but it sounds a littlebit wired to me.

app/models.py
```diff
@@ -596,6 +605,56 @@ def add(self, domain_name, domain_type, soa_edit_api, domain_ns=[], domain_maste
...
+            if 0 != len(domain_users):
+                self.name = domain_reverse_name
+                self.grant_privielges(domain_users)
+                return {'status': 'ok', 'msg': 'New reverse lookup domain created with granted privilages'}
+            return {'status': 'ok', 'msg': 'New reverse lookup domain created without users'}
+        return {'status': 'ok', 'msg': 'Reverse lookup domain already exists'}
+
+    def get_reverse_domain_name(self, reverse_host_address):
+        if re.search('ip6.arpa', reverse_host_address):
+            for i in range(31,3,-1):
```
petersipos: @olfway here happens the 'magic' :D

olfway: ah, i get it
for ipv4 you checks for
192.in-addr.arpa
168.192.in-addr.arpa
39.168.192.in-addr.arpa

that's what i need, thanks! i do not want to have "in-addr.arpa" zone

why not to check in reverse order? ie
39.168.192.in-addr.arpa
168.192.in-addr.arpa
192.in-addr.arpa

this way you will be able to have one or several special X.168.192.in-addr.arpa zones and use 168.192.in-addr.arpa zone for all other networks

jallakim: More-specific first would be best, yes -- especially for IPv6 (you could have a /32 catch-all, but still have more-specific /64s or similar).

petersipos: So, for example, I have a host with
`fe80::20c:29ff:fecb:7f31` IPv6 address and with
`1.3.f.7.b.c.e.f.f.f.9.2.c.0.2.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.e.f.ip6.arpa` reverse address, then I need to check the existance of:
`3.f.7.b.c.e.f.f.f.9.2.c.0.2.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.e.f.ip6.arpa` first and then
`f.7.b.c.e.f.f.f.9.2.c.0.2.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.e.f.ip6.arpa`
`7.b.c.e.f.f.f.9.2.c.0.2.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.e.f.ip6.arpa`
...
`f.ip6.arpa`
and if I don't find any of above I'll create the
`3.f.7.b.c.e.f.f.f.9.2.c.0.2.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.e.f.ip6.arpa` zone and add the
`1` PTR record ?

And the case of IPv4 addresses I whould need to do the same method?

olfway: In case of IPv4 - yes, same method
if you have host with 10.1.2.3 then you need to check existence of
2.1.10.in-addr.arpa first and then
1.10.in-addr.arpa
10.in-addr.arpa
and if not found then create "2.1.10.in-addr.arpa" zone and add "3" ptr record

For IPv6 i'm not sure what domain name you should create, maybe @jallakim can suggest

petersipos:
The PR is ready :)

petersipos:
@ngoduykhanh hello! What do you think about this PR? Did you have time to overlook this? :) Thank for reply ;)

ngoduykhanh:
@petersipos : Sorry for the delay, I didn't have time to look at this. I will review your PR. Thanks a lot.

petersipos:
Only I inquired :) non urgent. Thanks.

ngoduykhanh:
@petersipos : Your changes look good to me. Merging. Thanks for this awesome PR!

PR #181, opened 2017-03-01
Fix root domain record update via dyndns with ipv6 support. #113

joniw:
Hello,

i just expanded 7bcdc3cfd53c7e1ff186fca5775ea5289ea34d20 to support ipv6 addresses. Additionally the ip address is detected when it is malformed.

With regards

Jonathan


ngoduykhanh:
Fixed in #150.

PR #186, opened 2017-03-25
Add IPv6 support for DynDNS

afrimberger:


denschub:
@ngoduykhanh, do you have plans on merging this any time soon? :)

JiiPee74:
Ipv6 is not used everywhere so if you gonna add ipv6 support, there need to be a way to pull also ipv4 or things will break

afrimberger:
Seriously?
The implementation creates an A record for ipv4 and an AAAA record for ipv6 by detecting the ip address format / version. Therefore both versions are supported.

JiiPee74:
Well if I understood correctly, we are talking about dyndns service right?

So you setup your box to allow ssh connections in so you can connect your box from anywhere, but you have dynamic dns what does change time to time. (Depends of ISP alot)
So you setup dyndns service like this, good! Now we have hostname what we can use to connect.
I'm not a coder, but for me it looks like this will just set A or AAAA record, depends what your box happens to use to connect.
So if your box have ipv4/ipv6, but your dns server only use ipv4, your box ipv6 will never gets added right?
Or if both does have ipv6 support, AAAA record will be created. Now what happens when you try to ssh with your nice dyn.host.name from some box what does NOT support ipv6?
Yeah it won't work because dns server have just AAAA record from your box, not A record.

So at least documentation need some update also with some examples:
```
curl "http://user:pass@host/nic/update?hostname=sub.domain.com&myip=1.1.1.1"
curl "http://user:pass@host/nic/update?hostname=sub.domain.com&myip=[1EE7:BEEF::C0FE]"
```

Something like that.. Or if using some client, gotta make sure that client does support sending both IP types or there will be problems.

And how does powerdns admin with patch behave if you do curl update like that? Will it add both IP's, or just remove A record and add AAAA record?

afrimberger:
In the current state the PR relies on the already existing behaviour of updating/creating DynDNS records. It works as follows:
* the record type created first wins and can't be changed afterwards. So, if an A record was created, setting an AAAA record for that hostname won't work (and vice versa)
* "Exists" checks for a `Record` only consider the hostname, not the type
That could make sense in some special cases. But in general it would be better to handle both type of records independently. I therefore updated the PR accordingly. Executing the mentioned `curl` commands now creates two type of records and updates them accordingly.

JiiPee74:
I was reading dyn.com api about this and it says this:

> Authentication in URL
> 
> For web-browsers or utility programs (fetch, curl, lwp-request) that can parse authentication section in URL.
> 
> https://{user}:{updater client key}@members.dyndns.org/v3/update?hostname={hostname}&myip={IP Address}
> **NOTE: This authentication method supports both IPv6 and IPv4 addresses. Use commas to separate multiple IP addresses in the myip field.**

So if this project does not support same, I think it would be good idea to add support. That would basicly solve all the issues I was pointing out.

DennisSc:
I second that at least some mention in the documentation is strongly needed (or I am dumb enough to not find it) - because, I just spent a couple hours to set up some DIY php script and db tables, in order to implement dyndns functionality - only to discover, just after having finished, that everything is already there in pDNS-Admin (and ready to use). D'oh. 💡

PR #198, opened 2017-05-16
Expose and make the account field editable for DNS

sandeepsomasekharan:


ngoduykhanh:
Any reason to expose field? Not really helpful to me. And adding Account column in the domain table will take space i think.

PR #202, opened 2017-06-06
Docker configuration with ENV

mterzo:
I've added an alpine docker container to and copied the source to it.  I converted config_template.py to config.py which now uses os.getenv to managing most of the settings, i skipped the allowed to edit.

I've over written the docker-compose, with an example of who you this would be used with miradb.

With the new os.env the config['key'] may exist, but is set to none.  I've updated 1 place in the code this was a problem with LDAP.


rhaamo:
Looks fine and works. However I'm getting issues with logout not working and login but probably not related to this PR.

mterzo:
I’ve got it pointing at an AD and works.

rhaamo:
Yeah I've seen a bit later than my issue wasn't one.
Except that it builds and run perfectly.

ngoduykhanh:
I am refactoring the code and merge new features to `developement` branch. For now i won't merge this. Will consider to use your solution after refactoring.
Thanks for the PR!

PR #205, opened 2017-06-30
Allow for rDNS subdelegation via NS records

FrankPetrilli:


FrankPetrilli:
https://github.com/ngoduykhanh/PowerDNS-Admin/pull/208 seems to do this the way I actually intended to rather than the way it came out. Closing.

PR #206, opened 2017-06-30
Do not update domains from API on every dashboard request

hackedd:
Currently, viewing the dashboard does a request to the PowerDNS API and updates all domains. With large numbers of domains, this can take a substantial amount of time.

With this change, the domain list is only refreshed when the user clicks the 'refresh' button.

### DIFF
--- a/app/templates/dashboard.html
+++ b/app/templates/dashboard.html
@@ -124,6 +124,9 @@
           <div class="box">
             <div class="box-header">
               <h3 class="box-title">Hosted Domains</h3>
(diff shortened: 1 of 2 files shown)


ngoduykhanh:
This is useful when we have a bunch of domains that slow down the dashboard loading speed. However, it is not convenience to manually press the button every time we want to update the list of domains in case small scale of the deployment.

I am not gonna merge it for now. Will consider another solution after refactoring the code.

Thanks for the PR!

PR #214, opened 2017-10-10
Add google-oauth

patito:
Google oauth


ngoduykhanh:
Merged into `development` branch and added some adjustment to handle case that local user can not be created.

Waiting to merge to master after code refactoring. 

Thanks for your contribution!

PR #216, opened 2017-10-20
Add access control for non-admin users

hackedd:
Before this change, regular users would not see any domains they don't have access to, but they could just change the name of the domain in the URL to get access.


ngoduykhanh:
It is better to use decorator to check. See commit 023941df182adad73a514a90c26f141cdcf2d9e4.

Thanks for your contribution!

PR #224, opened 2018-02-08
Adding zone template functionality

verdel:
As a basis, I used the @ivanfilippov code. I asked permission of the author of the code for this.

I added the ability to create a zone template, the ability to clone an existing zone in a template, and create a zone based on the template.


PretorianX:
I vote for this functionality

PR #225, opened 2018-02-09
Fix some issues with LDAP authorization

verdel:
Fix issue with LDAP search filter, add ability to work with non-ascii data from LDAP and add ability to customize filter in case of use with Active Directory


ngoduykhanh:
Merged into `development` branch. Thanks for your contribution!

Issue #209, opened 2017-07-05
LDAP binding using username


maysara:
Greetings,

This issue has both a suggestion of feature, but also a bug regarding using LDAP configs with LDAP_TYPE='ad'

currently I need to have an LDAP account to do the binding, and write down the password in plain text in the config file! this seems a common practice, I'm sure some are comfortable with, but since one is using LDAP account to login (apart from the binding account); I wonder if it is possible to allow a configuration to use those credentials to bind.

I have modified my setup to do precisely that by replacing:
-            l.simple_bind_s(LDAP_USERNAME, LDAP_PASSWORD)
with
+            l.simple_bind_s(self.username,self.password)

in app/models.py

allowing the bind to happen using the credentials passed by user. I cant think of (security) drawbacks of this if filter is properly configured, except for the (hypothetical)  case where a service is disabled by disabling the user that is used to do the binding on the LDAP server.

It would be great if this can become a feature (I can provide a cleaner patch for an option such as LDAP_USE_LOGIN_CREDENTIALS_FOR_BINDING or so)

On a relevant note, although authenticating against AD, but I used config LDAP_TYPE='ldap' and manually set LDAP_FILTER = '(objectcategory=person)'; as the LDAP_TYPE='ad' behaviour seems to be:
    searchFilter = "(&(objectcategory=person)(samaccountname=%s))" % self.username
as it seems to break the below config:

LDAP_USERNAMEFIELD = 'userPrincipalName'

when LDAP_TYPE='ad'

If I understand it correctly, I guess thats a bug that prevents use of a custom usernamefield when ldap type is set to ad.

PR #227, opened 2018-03-31
better LDAP support, allowing direct and seach binding(s); in addition to a revised GROUP_SECURITY

maysara:
Greetings,

I've added a direct binding, distinct from search binding, this basically allows us to use the supplied username/password to bind a AD instead of a fixed DN. still preserving backward compatibility with previous configs, allowing GROUP_SECURITY and  LDAP_ADMIN_GROUP and LDAP_USER_GROUP (based on ivanfilippov work).

This is also an attempt to fulfill issue [#209](https://github.com/ngoduykhanh/PowerDNS-Admin/issues/209)

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -148,7 +148,8 @@ class User(db.Model):
             l.set_option( ldap.OPT_DEBUG_LEVEL, 255 )
             l.protocol_version = ldap.VERSION3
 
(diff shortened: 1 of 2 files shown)


ngoduykhanh:
Merged into development branch and did some adjustment to work with python 3. Waiting to merge to master.

Thanks for your contribution!

PR #239, opened 2018-04-10
Domain cloning feature and minor wording/spelling changes

xTrekStorex:
**First off, I am no python dev so this is probably nowhere close to your standards. I'd appreciate any improvements.**

I wrote this feature almost a year ago for my local fork in Gitlab (since this repo was pretty much dead).
I know we now have the template feature which is similar but I often encounter domain registrations where a clone would be useful but also not worth the template creation for a one-time thing. Also creating templates takes more time than a direct clone. 
So I pushed my changes on here to maybe get it included upstream.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -728,7 +728,7 @@ class Domain(db.Model):
                 domain_users.append(tmp.username)
             if 0 != len(domain_users):
                 self.name = domain_reverse_name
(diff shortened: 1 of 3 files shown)


ngoduykhanh:
"Clone" feature can be replaced by Domain Template which already implemented.
Also, Thanks for reporting the typo :)

PR #242, opened 2018-04-12
Some fixes and improvements 

mind04:
- Cloning a zone to a template did not remove the domain part from the records
- Stripping of the domain string was not limited to the the domain part of a name
- Fix clear history (firefox)


ngoduykhanh:
LGTM. However can you switch from `%` to `.format()` in formatting string?
Thanks for the contribution!

ngoduykhanh:
Merged! Thank you.

PR #246, opened 2018-04-17
Format history date/time before display using browser timezone

chinkung:
Fetch UTC date/time directly from DB and display it could be confused user.

It would be better to format it using user's browser timezone before display it to the user.


ngoduykhanh:
LGTM! Thanks for the contribution.

PR #252, opened 2018-04-18
Add ALLOW_HISTORY_CLEAR configuration to control history clearing behavior

chinkung:
DNS administrator might not like to allow any people who got admin permission to clear the history log in order to keep track who commit the changes to their DNS system.

This new config uses to control history clearing behavior in history management tab
If this key is set to False, clear history log is prohibited and 'Clear History' button in History Management page will be hidden.

I set ALLOW_HISTORY_CLEAR = True in config file to preserve current behavior.

### DIFF
--- a/app/templates/admin_history.html
+++ b/app/templates/admin_history.html
@@ -20,11 +20,13 @@
                 <div class="box-header">
                     <h3 class="box-title">History Management</h3>
                 </div>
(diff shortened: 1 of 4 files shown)


ngoduykhanh:
Config is going to be managed in the DB instead of file. `Operator` role added. Only `Administrator` users can remove history for now. See PR #345

Issue #130, opened 2016-08-31
Allow users to create domains


brian4286:
I would like to request a feature to allow users access to create new domains. It would seem that any feature allowing end users to add domains should have some limits involved to keep them abusing the system such as number of domains or possibly number of records. I would also think that limiting the type of domain (Native, Master, Slave) would be important.

HLFH:
@jonsparks Great! Can you submit a PR when it's ready? Thank you so much!

robetus:
Looks like this never happened. Anyone else working on something like this? I'm still interested.

ngoduykhanh:
Interesting feature. I gonna work on this after merging all PR.

robetus:
Great @ngoduykhanh looking forward to it. You might want to check out @jonsparks work before you begin. It looks like he did do some work on it.

icedream:
I'm still interested in this. 👍 How's the progress?

ngoduykhanh:
I will work on this after finishing the configuration management and security part which discuss at #342. You guys can leave your thoughts there :).

nvmng:
Great work. Any plan to allow assigning a dns template to a specific user?

PR #255, opened 2018-05-02
Improve SAML support

icb-:
Accept IdP EntityID to use when metadata contains more than one IdP.
Allow specifying attribute names to get given name, surname, and email address.
Allow specifying NameIDFormat to request.
Allow specifying whether to get username from a named attribute, or NameID.
Allow getting administrator state from attribute.

All new config options default to their previous functionality if not set.


ngoduykhanh:
Thank you for your contribution. 
@thomasDOTde you might want to see this changes.

icb-:
Is there anything you want to see different here to merge?

icb-:
@ngoduykhanh If #130 is going to happen, would you rather see an attribute for "user role" than just a boolean for "administrator" that's currently in the PR?

Azylog:
Hello,
This feature would really be awesome. So far, PowerDNS-Admin have templating, multi-users, one-click DNSSEC activation, almost everything needed !
I don't really understand which conflicts should be resolved to integrate this into master. May I help ?

icb-:
@Azylog the only conflict I see is the imports in `views.py`. That's easy enough to resolve. Before merging, I would like to have support for managing which Account(s) a user belongs to from SAML assertions, which shouldn't be too difficult to add.

ngoduykhanh:
Merged as it has been verified. Thanks again for the PR!

PR #256, opened 2018-05-07
Fail when non-numeric otp token was inserted

rene-dekkers:
Please observe this (very small) change. It is needed to friendly fail with a message when a non-numeric otp was inserted.

Thanks!

Regards René Dekkers


ngoduykhanh:
Merging. Thanks for your contribution !

PR #260, opened 2018-05-17
Debug mode configurable

Byorun:
Debug mode was hardcoded to True in run.py. I added a variable in order to be able to toggle debug mode via config file.

### DIFF
--- a/config_template.py
+++ b/config_template.py
@@ -7,6 +7,7 @@ SECRET_KEY = 'We are the world'
 BIND_ADDRESS = '127.0.0.1'
 PORT = 9191
 LOGIN_TITLE = "PDNS"
+DEBUG_MODE = True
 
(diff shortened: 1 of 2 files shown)


ngoduykhanh:
Are you using `./run.py` in production mode? It was supposed to used only in development environment where we need Debug enabled.

For other environment we might uses difference http web server, gunicorn for example; then the debug mode will be handled in a different way.

Byorun:
right now I am using the pure ./run.py with nginx as a reverse proxy inside my virtual cluster.

ngoduykhanh:
Better not to use `./run.py` method. It uses the flask built-in web server which is not recommended for production environment. See http://flask.pocoo.org/docs/0.12/deploying/. 

You can also take a look at the repo wiki page for other methods

Issue #267, opened 2018-05-27
Please add a per-domain comment


tmuncks:
There are several times where a comment for a domain would be useful.

In my case, we handle a lot of different domains, for a lot of different customers. Having a per-domain comment (and having it visible in the dashboard), would make it much easier to track/remember which domain belongs to which customer, who at the customer site normally manages the domain etc.

The PowerDNS API allows a comment to be set per-record, and perhaps setting this comment on the SOA record would be a great way of storing that information for the domain.

What do you think?

tmuncks:
Having looked at the API a bit more closely, PowerDNS has a built-in per-domain `account` parameter, that would be perfect for this. It is stored with the domain in the database and designed for local use.

https://doc.powerdns.com/md/httpapi/api_spec/#zones - see parameters here.

ngoduykhanh:
Yes, this feature has been mentioned before but i still haven't had time for it yet. I will try to implement it when i have time :).

tmuncks:
I can try to do a PR for it...

tmuncks:
Pull request #269 fixes this in a slightly more elaborate way.

tmuncks:
@ngoduykhanh feel free to close this issue, unless you feel a different type of per-domain comment would make sense, in addition the Account stuff.
I can see the use case for a per domain "free text" comment, but my initial idea with this issue, was to keep track of domain ownership, which the Account stuff takes care of.

PR #269, opened 2018-06-02
Initial support for Accounts

tmuncks:
This adds initial support for accounts a concept meant to signify a customer, a department or any other entity that somehow owns or manages one or more domains.

Meant to solve #267 but using an approach that allows for greater control and easier management.

The purpose is to be able to assign an account to any number of domains, making it easy to track who owns or manages a domain, significantly improving manageability in setups with a large number of domains.

An account consists of a mandatory, unique `name` and optional `description`, `contact` name and `mail` address. The account `name` is stripped of spaces and symbols, and lower cased before getting stored in the database and in PowerDNS, to help ensure some type of predictability and uniqueness in the database.

The term *account* is actually taken from the PowerDNS database, where the `domains.account` column is used to store the account relationship, in in the form of the account `name`.

The link to a domain in PowerDNS-Admin is done through the `domain.account_id` FOREIGN KEY, that is linked to the `account.id` PRIMARY KEY.

**Important notices**
*Database usage*
As we utilize the `domains.account` column of PowerDNS to properly persist the account relationship in a domain, people using this column for other purposes should be careful. PowerDNS-Admin did not previously use this column at all, so the chance of this being an issue is probably relatively small. 

*Account associations*
If an account is deleted, the account association is actively removed from all it's domains, but the domains themselves will not be modified or deleted in any way. Re-adding an account with the same `name` later, *WILL NOT* restore the old associations associations.

However, if the PowerDNS-Admin database gets re-initialized for whatever reason, adding a new account with a `name` previously assigned to a domain *WILL* restore the association. This helps simplify recovery in case of data loss.

**Account access**
In setups where multiple groups of people are managing a lot of domains, managing access to domains through a simple account relationship will save a lot of time and help ensure access rights are correct.

The account concept has grown the possibility of managing domain access permissions through accounts to provide this exact feature. When creating or editing an account, it's possible to add users to the account, through the same familiar multi select control, used on the domains. Users added to an account, automatically gains access to every domain that is tied to that account. Conversely, removing a user from an account, will revoke his access to all domains tied to the account. 
*Note:* This process does not interfere with per-domain access at all, so a user might still have permission to a domain through per-domain permissions after removing him from an account.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -486,7 +486,7 @@ class Account(db.Model):
         """
         account = Account.query.filter(Account.name == account_name).first()
         if account is None:
-            return 0
+            return None
 
         return account.id
 
@@ -664,7 +664,7 @@ class Domain(db.Model):
     notified_serial = db.Column(db.Integer)
     last_check = db.Column(db.Integer)
     dnssec = db.Column(db.Integer)
-    account_id = db.Column(db.Integer, db.ForeignKey('account.id'), nullable = False)
+    account_id = db.Column(db.Integer, db.ForeignKey('account.id'))
     account = db.relationship("Account", back_populates="domains")
     settings = db.relationship('DomainSetting', back_populates='domain')
 
--- a/app/models.py
+++ b/app/models.py
@@ -355,25 +355,48 @@ class User(db.Model):
             db.session.rollback()
             return False
 
+    def get_account_query(self):
+        """
+        Get query for account to which the user is associated.
+        """
+        return db.session.query(Account) \
+            .outerjoin(AccountUser, Account.id==AccountUser.account_id) \
+            .filter(AccountUser.user_id==self.id)
+
+    def get_account(self):
+        """
+        Get all accounts to which the user is associated.
+        """
+        return self.get_account_query()
+
     def get_domain_query(self):
-        return db.session.query(User, DomainUser, Domain) \
-            .filter(User.id == self.id) \
-            .filter(User.id == DomainUser.user_id) \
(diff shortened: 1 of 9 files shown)


tmuncks:
Almost ready with the "allow users to be associated with an account an gain domain access through that" code and expect to be able to push it here within 24 hours. I'll update this PR.

tmuncks:
There - Account permissions added

tmuncks:
@ngoduykhanh having done development on sqlite3, there are a few changes needed to make this work on MySQL/MariaDB. Hang on...

tmuncks:
With the latest commit, this works well on my system... The db_update stuff is not included yet - I need to look into how that works.

rodehoed:
@tmuncks great work! You have an mysql alter table command so I can test this functionality?

tmuncks:
@rodehoed absolutely...

Something like this should work - pastebin'ed to keep markdown from messing with it too much :)

https://pastebin.com/sn57y0Cz

Depending on your DB backend, you may need to tweak it a bit, but it should be pretty clear what it's trying to achieve.

Please let me know if you stumble on anything.

FWIW, we're working on a good solution for the DB migration, so that part is on it's way.

ngoduykhanh:
After the discussion with @tmuncks, we are OK with the current implementation and workaround to fix the SQLite migration issue related to ALTER TABLE.

FYI: We are going to switch from current migration scripts to flask-migrate soon.

ngoduykhanh:
The new feature in this PR looks good to me. Merging it now. Thanks @tmuncks for the contribution.

PR #271, opened 2018-06-05
Update models.py

sinzee:
I changed for api url to use domain name but domain object name in function update_from_master.


ngoduykhanh:
LGTM. Thanks for the PR.

PR #272, opened 2018-06-06
Fix OTP validation

tmuncks:
The result from the form is never an int but rather a string of digits, so that's what we should be checking for.

This fixes OTP validation


ngoduykhanh:
LGTM, thansk for the PR.

PR #273, opened 2018-06-06
Restrict certain admin changes on the current user

tmuncks:
Disable the admin toggle and delete operations from the current user, to avoid accidents.


ngoduykhanh:
LGTM. Thanks for the improvement.

PR #274, opened 2018-06-06
Add config option to allow DNSSEC changes only for admins

tmuncks:
DNSSEC requires changes to the parent domain, which in many cases requires special access to a registry or the like.
For that reason, especially the option to disable DNSSEC can be dangerous - if DNSSEC is disabled in PowerDNS but not in the registry, the domain stops working.

For this reason, adding an option to disable DNSSEC changes for non-admins seems reasonable.


ngoduykhanh:
Merge your PR. Also did some improvement to check the in the flask route when user enable/disable dnssec. Commit 2958ae663cfbfaffd62e9d01ac4040d8c7b30351.

PR #277, opened 2018-06-10
Automatically rectify DNSSEC enabled zones

### LINKED ISSUE #268: DNSSEC-enabled zones should be rectified
Hey, i had some DNS issues after changing records via PowerDNS-Admin. I found out that PowerDNS-Admin does not rectify zones after updating. May you can add this?

See: https://doc.powerdns.com/authoritative/http-api/zone.html (at the bottom)

tmuncks:
For DNSSEC enabled zones to function correctly, they need to be rectified on update.

This changes the DNSSEC enable/disable code to set API-RECTIFY:

To `true` when activating DNSSEC on a domain
To `false` when deactivating DNSSEC on a domain

With this, PowerDNS promises to handle the needed rectifications.

Aims to fix #268 

Fixed a different problem in the process, where an exception is thrown (and caught) in the DNSSEC delete code, because utils.fetch_json() returns a BOOL on the DELETE request, rather than a json.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -842,19 +842,30 @@ class Domain(db.Model):
         if domain:
             headers = {}
             headers['X-API-Key'] = PDNS_API_KEY
-            post_data = {
-                "keytype": "ksk",
-                "active": True
-            }
             try:
+                # Enable API-RECTIFY for domain, BEFORE activating DNSSEC
+                post_data = {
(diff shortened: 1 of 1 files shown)


ngoduykhanh:
LGTM. Merging. Thanks for submitting the PR.

PR #280, opened 2018-06-11
Use webassets

kaechele:
This removes all bundled frontend libraries and replaces them with a yarn + webassets workflow.
The docker configuration has been adapted accordingly.

Note that this does increase the initial boot-up time of the application considerably, which is why we generate the assets before launching the application on the docker image.


ngoduykhanh:
Nice work @kaechele . Merged your PR.
For non-docker running, we should generate the assets before launching the PowerDNS-Application right? If so, I will update the wiki pages for the instruction.

kaechele:
Good Idea @ngoduykhanh 
While Flask-Asssets will build the assets on demand it will take about a minute for the first page to load if the assets are not built beforehand. That might irritate users.
So a `flask assets build` is advisable.

Either way, the user will need to run `yarn` before he's able to use the tool at all without docker. I guess at least that should be documented, also since it requires yarn to be installed.

PR #282, opened 2018-06-12
Enable pool_pre_ping in DB connection

tmuncks:
To avoid problems with inactive DB connections, SQLAlchemy provides a `pool_pre_ping` option, that is described in more detail here:

http://docs.sqlalchemy.org/en/latest/core/pooling.html#disconnect-handling-pessimistic

In Flask environments, it's enabled by subclassing SQLAlchemy, which is what I've done here.

This fixes errors like:
sqlalchemy.exc.OperationalError: (_mysql_exceptions.OperationalError) (2006, 'MySQL server has gone away') which results in an Error 500 in the UI.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,7 +1,15 @@
 from werkzeug.contrib.fixers import ProxyFix
 from flask import Flask, request, session, redirect, url_for
 from flask_login import LoginManager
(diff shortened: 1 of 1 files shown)


ngoduykhanh:
This option loos useful to me. Merged it. Thank you for submitting the PR.

Issue #194, opened 2017-04-30
NoneType' object has no attribute 'value'


madpsy:
After performing a 'git pull' (by accident I must admit), for some reason pdns-admin broke.

This occurs if if run directly (./run.py) or via wsgi. It was working beforehand so not sure what could have gone wrong here.

Anyone any thoughts?

`Traceback (most recent call last):
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/werkzeug/contrib/fixers.py", line 152, in __call__
    return self.app(environ, start_response)
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/flask_login/utils.py", line 227, in decorated_view
    return current_app.login_manager.unauthorized()
  File "/var/www/pdns-admin/flask/lib/python2.7/site-packages/flask_login/login_manager.py", line 148, in unauthorized
    return self.unauthorized_callback()
  File "/var/www/pdns-admin/app/views.py", line 201, in login
    basic_enabled=BASIC_ENABLED, signup_enabled=SIGNUP_ENABLED)
[...]

Issue #175, opened 2017-01-10
PowerDNS-Admin fails to start after update due to auto_ptr


Olipro:
It appears to me that there is no DB migration for the auto_ptr - or at least, running the DB upgrade script did nothing.

Manually adding the auto_ptr field to the setting table fixed it. I'm using PostgreSQL.

PR #287, opened 2018-06-22
Improve setting handling

tmuncks:
Moves setting definitions into code (rather than database).

Aims to fix issues such as #175 and #194 .

For a setting to be useful, the code has to be able to make sense of it anyway. For this reason it makes sense, that the available settings are defined within the code, rather than in the database, where a missing row has previously caused problems. Instead, settings are now written to the database, when they are changed.

So instead of relying on the database initialization process to create all available settings for us in the database, the supported settings and their defaults are now in a `defaults` dict in the Setting class. With this in place, we can stop populating the `setting` table as a part of database initialization and it will be much easier to support new settings in the future (we no longer need to do anything to the database, to achieve that).

Another benefit is that any changes to default values will take effect automatically, unless the admin has already modified that setting to his/her liking.

To make it easier to get the value of a setting, falling back to defaults etc, a new function `get` has been added to the Setting class. Call it as `Setting().get('setting_name'), and it will take care of returning a setting from the database or return the default value for that setting, if nothing was found.

The `get` function returns `None`, if the setting passed to the function, does not exist in the `Setting.defaults` dict - Indicating that we don't know of a setting by that name.

---

I've thrown in a simple typo fix to the `set_maintenance` (`set_mainteance`) function. It's a simple change and it would conflict with the setting handling commit here, if I had PR'ed it on it's own. Hopefully this works for you.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -884,8 +884,7 @@ class Domain(db.Model):
         domain_obj = Domain.query.filter(Domain.name == domain_name).first()
         domain_auto_ptr = DomainSetting.query.filter(DomainSetting.domain == domain_obj).filter(DomainSetting.setting == 'auto_ptr').first()
         domain_auto_ptr = strtobool(domain_auto_ptr.value) if domain_auto_ptr else False
-        system_auto_ptr = Setting.query.filter(Setting.name == 'auto_ptr').first()
-        system_auto_ptr = strtobool(system_auto_ptr.value)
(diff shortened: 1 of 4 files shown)


tmuncks:
***Note***: These same changes are in PR #289 too. #289 solves a different issue, using the updated setting infrastructure proposed here.

ngoduykhanh:
LGTM. Merging the PR. Thanks for the contribution @tmuncks

PR #293, opened 2018-06-27
Wait for MySQL to be ready before initialising/upgrading database when starting container

Sam-Martin:
Adds a healthcheck on container start to ensure the MySQL DB is ready.
This isn't an issue in the docker-compose because of the way `depends_on` works in v2.1, but when upgrading to newer versions (or orchestrating user a different tool), this becomes an issue when deploying DB and app simultaneously.


ngoduykhanh:
Ok, this might helpful in case we deploy on other environment. Merged it. Thanks for submitting the PR.

PR #301, opened 2018-07-03
Fix SQLAlchemy errors when creating LDAP users

an-empty-string:
When creating new users from LDAP, SQLAlchemy will throw a parameter binding exception because the values returned from LDAP lookups are `bytes`, not `str`s like SQLAlchemy expects. This fixes that by `.decode()`ing the values returned from LDAP before trying to do anything with them.

Note that the actual lookup doesn't cause an exception, it's only triggered when actually inserting the user into the database.

This might (probably will) break things when running on Python 2; I'm not sure how much you care about that.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -256,9 +256,9 @@ class User(db.Model):
                 try:
                     # try to get user's firstname & lastname from LDAP
                     # this might be changed in the future
-                    self.firstname = result[0][0][1]['givenName']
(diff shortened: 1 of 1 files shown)


an-empty-string:
Wait, no, looks like this actually causes other bugs on my instance. I'll try to figure out the root cause and resubmit this PR later.

PR #312, opened 2018-07-27
Use the dictionary get() method instead of directly accessing a proba…

Finn10111:
…bly not existing key to prevent a KeyError while calling the URL /dashboard-domains-updater - Error message was: Can not update domain table. Error: 'account'

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -759,7 +759,7 @@ class Domain(db.Model):
 
             # update/add new domain
             for data in jdata:
-                account_id = Account().get_id_by_name(data['account'])
+                account_id = Account().get_id_by_name(data.get('account'))
                 d = dict_db_domain.get(data['name'].rstrip('.'), None)
                 changed = False
                 if d:


Finn10111:
I am using PowerDNS 3.4.1 running on Debian 8 and all the entries in the `domains` table look like this: 

```
 id |        name         | master | last_check |  type  | notified_serial | account 
 10 | example.com         |        |            | MASTER |      1532709027 | 
```
With your last sentence you figured out the problem, the `account` column is empty for all domains. But if I understood the PowerDNS documentation correctly, the `account` column is not mandatory. At the moment it seems to be mandatory if someone wants to use PowerDNS-Admin.

> There is no need to fill the account name out but it does help keep track of where a domain comes from.

https://doc.powerdns.com/md/authoritative/modes-of-operation/

I only have a small amount of domains so I do not need the account. There are probably more scenarios where someone does not set the account, what do you think?

tmuncks:
On PowerDNS 4.x, the API always populates `account` key, regardless of the contents of that column. It doesn't matter if the column is empty, populated or even `NULL`, it will always be in the API result:

```
+----+-------------+--------+------------+--------+-----------------+----------+
| id | name        | master | last_check | type   | notified_serial | account  |
+----+-------------+--------+------------+--------+-----------------+----------+
|  9 | example.org |        |       NULL | NATIVE |            NULL |          |
| 10 | ttt.xyz     |        |       NULL | NATIVE |            NULL | NULL     |
| 11 | tets.gl     |        |       NULL | NATIVE |            NULL | someacct |
+----+-------------+--------+------------+--------+-----------------+----------+
```

```
$ curl -s -H 'X-API-Key: secret' http://127.0.0.1:8081/api/v1/servers/localhost/zones |jq
[
  {
    "account": "",
    "dnssec": false,
    "id": "example.org.",
    "kind": "Native",
    "last_check": 0,
    "masters": [],
    "name": "example.org.",
    "notified_serial": 0,
    "serial": 2018060601,
    "url": "/api/v1/servers/localhost/zones/example.org."
  },
  {
    "account": "",
    "dnssec": false,
    "id": "ttt.xyz.",
    "kind": "Native",
    "last_check": 0,
    "masters": [],
    "name": "ttt.xyz.",
    "notified_serial": 0,
    "serial": 0,
    "url": "/api/v1/servers/localhost/zones/ttt.xyz."
  },
  {
    "account": "someacct",
    "dnssec": false,
    "id": "tets.gl.",
    "kind": "Native",
    "last_check": 0,
    "masters": [],
    "name": "tets.gl.",
    "notified_serial": 0,
    "serial": 2018061001,
    "url": "/api/v1/servers/localhost/zones/tets.gl."
  }
]
```

So how does this look on 3.x? Is the account stuff even supported in the API? If you have something in the `account` column, does it appear in the API?

Finn10111:
@tmuncks You are probably right, the API in 3.1 does not seem to support the `account` column: 

```
 id |        name         | master | last_check |  type  | notified_serial |   account   
  3 | example.de          |        |            | MASTER |      2018072601 | 
 10 | example.com         |        |            | MASTER |      1532709027 | testaccount
```


`curl -s -H 'X-API-Key: secret' http://127.0.0.1:8081/servers/localhost/zones`

```
[
   {
      "id":"example.de.",
      "url":"/servers/localhost/zones/example.de.",
      "name":"example.de",
      "kind":"Master",
      "dnssec":true,
      "masters":[

      ],
      "serial":2018072601,
      "notified_serial":2018072601,
      "last_check":0
   },
   {
      "id":"example.com.",
      "url":"/servers/localhost/zones/example.com.",
      "name":"example.com",
      "kind":"Master",
      "dnssec":false,
      "masters":[

      ],
      "serial":1532709027,
      "notified_serial":1532709027,
      "last_check":0
   }
]
```

I know that PowerDNS 3.x is not supported anymore but it while take until I can upgrade to 4.1.

tmuncks:
Hmm, so I think we may need other changes as well, to avoid problems in other code paths due to this.

Let me run a few test tonight - I think we may need a few changes apart from this, to avoid problems.

Do you get similar errors when assigning an account to a domain (even though to might never want to do that personally, it makes sense to fix it properly)?

Finn10111:
I have tried your suggested solution and it works fine, also the log shows the following as expected: 

`[DEBUG] No 'account' data found in API result - Unsupported PowerDNS version?`

PR #313, opened 2018-07-30
Tolerate pdns 3.x API deficiencies

tmuncks:
PowerDNS 3.x API does not support setting or getting account info.

This patch lets PowerDNS 3.x users use the rest of the interface without problems.
Account stuff still does not work.

A message is logged in debug mode, to help with troubleshooting on newer versions
of PowerDNS.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -486,6 +486,10 @@ class Account(db.Model):
         """
         Convert account_name to account_id
         """
+        # Skip actual database lookup for empty queries
+        if account_name is None or account_name == "":
+            return None
+
         account = Account.query.filter(Account.name == account_name).first()
         if account is None:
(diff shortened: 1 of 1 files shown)


ngoduykhanh:
`account` field was introduced in version 3.4.3. Your change should fix for #312 . Merging. Thanks for submitting t he PR.

just FYI: There was several changes in pdns db schema since they upgraded from 3.x to 4.x. At the beginning i use [NEW_SCHEMA](https://github.com/ngoduykhanh/PowerDNS-Admin/blob/master/app/models.py#L64) flag to switch the code between 2 versions. Not sure if we have a better way to organize it as new changes in the pdns api might come in the future.

Issue #322, opened 2018-08-06
SAML  Invalid dict settings: idp_sso_not_found error


rodehoed:
Hi,

With PR #255 active i'm getting an cryptic exception:

```
Traceback (most recent call last):
  File "<path>VirtualENV/lib/python3.5/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "<path>VirtualENV/lib/python3.5/site-packages/werkzeug/contrib/fixers.py", line 152, in __call__
    return self.app(environ, start_response)
  File "<path>VirtualENV/lib/python3.5/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "<path>VirtualENV/lib/python3.5/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "<path>VirtualENV/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "<path>VirtualENV/lib/python3.5/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "<path>VirtualENV/lib/python3.5/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "<path>VirtualENV/lib/python3.5/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "<path>VirtualENV/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "<path>VirtualENV/lib/python3.5/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "<path>VirtualENV/lib/python3.5/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "<path>src/app/views.py", line 192, in saml_login
    auth = utils.init_saml_auth(req)
  File "<path>src/app/lib/utils.py", line 280, in init_saml_auth
    auth = OneLogin_Saml2_Auth(req, settings)
  File "<path>VirtualENV/lib/python3.5/site-packages/onelogin/saml2/auth.py", line 54, in __init__
    self.__settings = OneLogin_Saml2_Settings(old_settings, custom_base_path)
[...]

PR #329, opened 2018-08-11
Allow specifying SAML2 SSO binding format.

icb-:
Fixes ngoduykhanh/PowerDNS-Admin#322

Adds new optional config setting `SAML_IDP_SSO_BINDING` for which [SAML binding format](https://en.wikipedia.org/wiki/SAML_2.0#SAML_2.0_Bindings) to use. If not specified, use library default.


ngoduykhanh:
Yes, it happens sometimes with Travis CI. I restarted the test and merged your PR. Thanks for your contribution @icb- !

PR #330, opened 2018-08-12
Add option to edit users from the comfort of the UI

### LINKED ISSUE #325: Allow admins to modify users
We're missing the possibility for admins to edit users, change passwords, reset OTP token etc from the UI...

I'm working on a PR - expect a PR In a few days

tmuncks:
Fixes #325 

Update user management feature to allow editing user details directly in the admin user interface.

Also added an option to reset the two factor authentication data of a user, for when that's needed (lost device, technical issues etc).

### DIFF
--- a/app/templates/admin_edituser.html
+++ b/app/templates/admin_edituser.html
@@ -88,7 +88,7 @@
                 <p><strong>Beware: This could compromise security!</strong></p>
             </div>
             <div class="box-footer">
-                <button type="button" class="btn btn-flat btn-warning button_otp_disable" id="{{ user.username }}" {% if user.otp_secret == "" %}disabled{% endif %}>Disable Two Factor Authentication</button>
+                <button type="button" class="btn btn-flat btn-warning button_otp_disable" id="{{ user.username }}" {% if not user.otp_secret %}disabled{% endif %}>Disable Two Factor Authentication</button>
             </div>
         </div>
         {% endif %}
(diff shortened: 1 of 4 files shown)


app/templates/admin_edituser.html
```diff
@@ -72,22 +73,81 @@ <h4><i class="icon fa fa-ban"></i> Error!</h4>
...
+            <div class="box-header with-border">
+                <h3 class="box-title">Two Factor Authentication</h3>
+            </div>
+            <div class="box-body">
+                <p>If two factor authentication was configured and is causing problems due to a lost device or technical issue, it can be disabled here.</p>
+                <p>The user will need to reconfigure two factor authentication, to re-enable it.</p>
+                <p><strong>Beware: This could compromise security!</strong></p>
+            </div>
+            <div class="box-footer">
+                <button type="button" class="btn btn-flat btn-warning button_otp_disable" id="{{ user.username }}" {% if user.otp_secret == "" %}disabled{% endif %}>Disable Two Factor Authentication</button>
```
ngoduykhanh: There is the case that otp_secret = NULL when new user created. I think we should check with `if not user.otp_secret`.

tmuncks: Well spotted. I'll fix this...

ngoduykhanh:
Merging, thanks for submitting the new feature.

PR #331, opened 2018-08-12
Fix double refresh when activating 2FA

tmuncks:
When toggling Two Factor Authentication, it often takes a few tries to get it to work.
The toggle function ends up reloading the page in two different places, effectively creating a race condition.

This fixes that problem


ngoduykhanh:
Merged. Thanks @tmuncks

Issue #337, opened 2018-08-21
Issues with AD LDAP and ADFS


MaxDiOrio:
When I try to log in with a user using AD LDAP, I get a red box: 'LDAP_ADMIN_USERNAME' and nothing happens.   What I really want is User and Admin groups assigned by LDAP group membership.  Do the accounts need to be pre-created, or will it take care of creating the accounts?

LDAP_ENABLED = True
## AD CONFIG
LDAP_TYPE = 'ad'
LDAP_URI = 'ldap://la-2pdom01:389'
LDAP_USERNAME = 'CN=SVC_LDAP_RO,OU=General,OU=Service,OU=Accounts,DC=internal,DC=domain,DC=com'
LDAP_PASSWORD = 'password'
LDAP_SEARCH_BASE = 'DC=internal,DC=domain,DC=com'
## You may prefer 'userPrincipalName' instead
LDAP_USERNAMEFIELD = 'sAMAccountName'
## AD Group that you would like to have accesss to web app
#LDAP_FILTER = 'CN=PowerDNS_Users,OU=Infrastructure,OU=SecurityGroups,DC=internal,DC=domain,DC=com'
LDAP_FILTER = '(objectClass=user)'
LDAP_GROUP_SECURITY = True # True or False
LDAP_ADMIN_GROUP = 'CN=PowerDNS_Admins,OU=Infrastructure,OU=SecurityGroups,DC=internal,DC=ieeeglobalspec,DC=com'
LDAP_USER_GROUP = 'CN=PowerDNS_Users,OU=Infrastructure,OU=SecurityGroups,DC=internal,DC=ieeeglobalspec,DC=com'

ALSO
Does anyone have a clue how to get ADFS working properly with this?  We really need a WIKI article, and by rights, LDAP and ADFS should be integrated into the admin UI configuration with test buttons.

MaxDiOrio:
I'll tag on to this - I think I have it almost working properly now.  Turns out your config-template.py appears to be wrong.  The app is looking for LDAP_ADMIN_USERNAME as the attribute to search the ldap directory.  The config-template.py doesn't have this variable in there, it uses LDAP_USERNAME.  I copied the template out of development.py and am using that with appropriate changes.

[...]

Issue #335, opened 2018-08-17
AD integration


tavaresrick:
Hello, I'm trying to integrate the app with my local AD services.

I've created an group on the AD and I want all users in this group to have access to the application and automatically became Admins, my current config is this:

# AD CONFIG
LDAP_TYPE = 'ad'
LDAP_URI = 'ldaps://foo.corp:636'
LDAP_USERNAME = 'cn=svc_powerdns,ou=Services_TI,ou=sao_paulo,ou=usuarios,dc=foo,dc=corp'
LDAP_PASSWORD = 'p@ssword'
LDAP_SEARCH_BASE = 'ou=usuarios,dc=foo,dc=corp'
# You may prefer 'userPrincipalName' instead
LDAP_USERNAMEFIELD = 'sAMAccountName'
# AD Group that you would like to have accesss to web app
LDAP_FILTER = '(memberof=cn=gs_powerdns_admin,ou=Grupos,ou=sao_paulo,ou=usuarios,dc=foo,dc=corp)'

Is it possible to do that?

ngoduykhanh:
Let's discuss on #337 . Closing this one.

Issue #237, opened 2018-04-09
Profile edit deletes OTP secret and writes LDAP(?) password to DB


xTrekStorex:
When I edit my profile (on any user) it overwrites the OTP secret in the DB with an empty string.
This causes two-factor to consider any input valid, even no input at all.

![image](https://user-images.githubusercontent.com/3179810/38485149-1f05bfc2-3bd9-11e8-9fb1-ae77d9e61528.png)

So e.g. changing my name or email clears my two-factor setup. 
Also since in the DB it's changed to an empty string instead of NULL the "Enable two-factor" checkbox is unresponsive until I logout and log back in again. Then I can generate a new OTP QR code and scan it in FreeOTP. After that two-factor works again as expected. Until I change something on my profile again.

Also is it intended that passwords of LDAP users are written to the user table when a change like email is submitted - no input in the password tab? At least I think it's that password, could be some random hash though.
But shouldn't LDAP users be handled as external users as well, disabling the password tab like with SAML auth (94f8dd84fbab064ea8e2da87b930785951fb3662 & 05e24e5e46cf4a29d49304624aa5f6938beef6f3)?

![image](https://user-images.githubusercontent.com/3179810/38485352-b72cf98c-3bd9-11e8-8947-00f3170a19d9.png)

ngoduykhanh:
Fixed in commit 84d4bfa. Can you double check?

xTrekStorex:
Now I can't change anything on LDAP users (including avatar) and can't setup or disable two-factor.
Read-only name and email are fine, but shouldn't 2FA still be available through PDNSAdmin even for external auth? I hadn't looked at the OTP code section in the aforementioned commits so I didn't expect that. 

![image](https://user-images.githubusercontent.com/3179810/38540214-7790d0e2-3c9b-11e8-96fe-815d75b5d400.png)

After logging in with the new code changes all OTP secrets were cleared btw
Also there are still password hashes written to the DB at login (I flushed them via SQL before). 
Or are those just salted empty strings? I'd feel a bit uncomfortable if those were actual hashed passwords pulled from LDAP

![image](https://user-images.githubusercontent.com/3179810/38542555-66bfbe52-3ca2-11e8-8673-4422275d0fe2.png)

ngoduykhanh:
[...]

PR #339, opened 2018-08-22
Reading app config from DB and authentication adjustment

ngoduykhanh:
This PR is to:
- Move the configuration from file to database.
- Fix AD group security authentication #335 #337
- Enable OTP feature for LDAP user #237


ngoduykhanh:
TODO: Manage SAML config in DB also.

PR #352, opened 2018-09-02
Allow CNAMEs to point to '@'

tmuncks:
Often, you need a CNAME to point to the base domain name, but with a lot of domains and a lot of records, it can become tedious and even error prone to have to type out the domain name in full.

This PR enables the use of the shorthand `@` (or blank) for `CNAME` records, so that if a CNAME record is specified to point to '@' or '', it's replaced with the base domain name.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1346,10 +1346,12 @@ class Record(object):
                         if ':' in r['record_name']: # dirty ipv6 check
                             r_name = r['record_name']
 
+            r_data = domain if r_type == 'CNAME' and r['record_data'] in ['@', ''] else r['record_data']
+            
             record = {
                         "name": r_name,
                         "type": r_type,
-                        "content": r['record_data'],
+                        "content": r_data,
                         "disabled": True if r['record_status'] == 'Disabled' else False,
                         "ttl": int(r['record_ttl']) if r['record_ttl'] else 3600,
                     }


ngoduykhanh:
Hi @tmuncks , why do we need a CNAME record for the base domain? According to RFC1912, CNAME records can only exist as single records and can not be combined with any other resource record. I haven't ever seen other domain provider allow to use CNAME for the root domain either.

tmuncks:
Hi @ngoduykhanh, I'm sure I'm not explaining myself clearly enough - My apologies.

Consider the following zone-snippet (127.0.0.1 obviously used only as an example):
```
$ORIGIN superlongandverydifficulttotypedomainname.com.
@     IN A        127.0.0.1
www   IN CNAME    superlongandverydifficulttotypedomainname.com.
test  IN CNAME    superlongandverydifficulttotypedomainname.com.
dev   IN CNAME    superlongandverydifficulttotypedomainname.com.
user  IN CNAME    superlongandverydifficulttotypedomainname.com.
my    IN CNAME    superlongandverydifficulttotypedomainname.com.
```

Now if instead of having to type in the full domain name for each of these CNAMEs, that would come in handy:
```
$ORIGIN superlongandverydifficulttotypedomainname.com.
@     IN A        127.0.0.1
www   IN CNAME    @
test  IN CNAME    @
dev   IN CNAME    @
user  IN CNAME    @
my    IN CNAME    @
```

Also - And perhaps more evidently useful actually - This extends to templates, where it's now possible to create this kind of record up front, having it point to the actual name of the newly created domain.

Does this clear things up?

ngoduykhanh:
Hi @tmuncks ,

Thanks for the detail explanation. I misunderstood your idea. Sorry for that. The changes looks good to me. Merging the PR...

mchlrv:
This doesn't seem to work (anymore). If I create a template with an CNAME that points to an '@', it will end up like this in the PowerDNS database.

PR #353, opened 2018-09-02
Change user-dropdown button colors

### LINKED ISSUE #332: [RFE] Logout button not easily visible
I spent sometime and finally found the logout button.

image

Is it possible to have both buttons in color other than existing white ?

tmuncks:
Making it a little more evident, that the profile- and logout buttons exist in that dropdown.

Aiming to fix #332 

screenshot from 2018-09-02 12-29-58

Colors are taking the elsewhere in the theme and - to me - seems to match.

### DIFF
--- a/app/static/custom/css/custom.css
+++ b/app/static/custom/css/custom.css
@@ -32,3 +32,6 @@ table td {
    color: olive;
 }
 
+.user-footer {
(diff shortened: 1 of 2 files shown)


ngoduykhanh:
LGTM, merging it. Thanks @tmuncks .

PR #377, opened 2018-10-01
Setting a session lifetime

odumasFR:
This patch is meant to have a session timeout shorter than the default flask_login one.
I manually added a new setting called session_timeout in the setting table, with a default value of 10 minutes:
insert into setting(name,value, view) values('session_timeout', 10, 'basic');

I didn't quite understand the way new settings were added in the database depending on the version of alembic (files in migrations/versions), It may be necessary to complement this patch with the proper migration file.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1808,6 +1808,7 @@ class Setting(db.Model):
         'allow_user_create_domain': False,
         'bg_domain_updates': False,
         'site_name': 'PowerDNS-Admin',
+        'session_timeout': 10,
         'pdns_api_url': '',
         'pdns_api_key': '',
         'pdns_version': '4.1.1',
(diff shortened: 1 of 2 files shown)


ngoduykhanh:
If you decided to manage this config in the DB, i suggest you to put it _Setting > Basic_ view which requires to adjust [this list](https://github.com/ngoduykhanh/PowerDNS-Admin/blob/master/app/views.py#L1360) also.

You don't need to manually insert the record for that config, and you shouldn't do that. Without available setting in the DB, PDA should work. It will take the default value in the [defined dictionary](https://github.com/ngoduykhanh/PowerDNS-Admin/blob/master/app/models.py#L1797).

odumasFR:
I modified the list you mentioned. Since the database insert wasn't in any commit, there is no other  needed modification.

ngoduykhanh:
LGMT. Merging it. Thanks for submitting the PR.

PR #391, opened 2018-11-06
support arbitrary db port

deathowl:
expose environment variable, make entrypoint use the newly added env var.


ngoduykhanh:
Hi,

Thanks for submitting the PR. I see you introduced new variables which are `PDNS_PROTO` and  `PDNS_PORT `. Where should those variables come from?

Also, can you rebase into single commit for this PR?

Thanks.

deathowl:
PDNS_PORT and PDNS_PROTO can be supplied via environment variables, but are not required to be, since they fall back to the default vaules

ngoduykhanh:
LGTM now. Merging the PR. Thanks @deathowl .

PR #406, opened 2018-11-24
Make a redirection to previous page after logged in

### LINKED ISSUE #403: Login fails with "Method not allowed" on some URLs
Hi,

When accessing on some routes ('/domain/<path:domain_name>', '/template/<path:template>/edit, ....) and session is expired, the login view is displayed without redirection to "/login" route.
So far so good.
But when you get logged in, these route do not allow POST method so error occurs.

Two way to fix that :
1. Allow POST method on each route which is protected
    - PROS : easy to fix
    - CONS : the current redirection to /dashboard lose the user current route
2. Make a real redirection to /login route with a **next** param to get redirected on the previous route after logged in
    - PROS : a better user experience I think
    - CONS : need to change the login view and add redirection on protected views

Any hints ?

ymage:
Fix #403

### DIFF
--- a/app/views.py
+++ b/app/views.py
@@ -299,7 +299,6 @@ def saml_authorized():
         return  render_template('errors/SAML.html', errors=errors)
 
 
-@app.route('/login', methods=['GET', 'POST'])
 @login_manager.unauthorized_handler
 def unauthorized_callback():
     session['next'] = request.path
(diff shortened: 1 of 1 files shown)


ngoduykhanh:
LGTM. Merging the PR. Thanks for your contribution!

PR #409, opened 2018-11-24
Add ttl_options setting

### LINKED ISSUE #362: [Request] Custom TTL (Or 10m/2h/6h/12h option)
Hi,
The TTL of 1m/5m/15m/30m/60m/24h is almost perfect for 99%

But between 1h and 24h there is a big gap xD
Ans a 10m option would be nice too :-)
Or maybe an setting for custimze that time ranges. But thats not important, just a nice to have :-)

And still looking forward for the PTR Tab xD
because IPv6 entrys makes the whole dashboard ugly :-)
Cheers :-)

rkerr:
Adds a new setting to define the options in the TTL dropdown when
editing a record. The setting is a comma separated string with the
valid options.

Fixes #362

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -12,6 +12,7 @@ import dns.reversename
 import dns.inet
 import dns.name
 import logging as logger
+import pytimeparse
 
 from ast import literal_eval
 from datetime import datetime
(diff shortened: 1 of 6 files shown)


ngoduykhanh:
Thanks @rkerr for the contribution.  Your changes look good to me. However, it should be applied to domain template also.

There will be also a small issue if we remove the TTL setting that already applied to an existing record. User will see the blank field when editing it. But i think we can ignore it for now.

ngoduykhanh:
Yes, your solution looks good to me. Merging the PR. Thanks for your contribution!

PR #412, opened 2018-11-26
Fix login redirection when submounted

### LINKED ISSUE #411: Login redirection does not work when sub-mounting app (eg wsgi)
When sub-mounting app with a prefix for Flask route, login redirects to root path URL.

Refer to #403

ymage:
Fix #411 to use with APPLICATION_ROOT Flask config variable


ngoduykhanh:
Merged it. Thanks for the patch.

PR #419, opened 2018-12-05
Added Config to address issue 122

sristeter:
Added config option to enable/disable display of otp token field on login page.


ymage:
Thanks for that :)

ngoduykhanh:
Hi, Thanks for submitting the PR. It looks like this config is for controlling the UI only. What will happen if a user had OTP enabled but we disable OTP field it in the UI. They can't login?

PR #448, opened 2019-01-24
Better LDAP authentication settings help text

nickbouwhuis:
In the current help text for configuring LDAP authentication, it isn't really clear that one should include the parentheses. This should help prevent this simple error, helping people save time in the future.


ngoduykhanh:
Adding double quote for those filter doesn't work for me. LDAP server complained `ldap.FILTER_ERROR: {'desc': 'Bad search filter'}`. 
Why do you need that to make it work? Is it related to the value stored in the DB, python version or something else ?

PR #455, opened 2019-02-02
dyndns: accept and validate both A and AAAA records; default to client address

mtdcr:
Besides accepting IPv6 addresses in general, this implements two items from https://help.dyn.com/remote-access-api/perform-update/ regarding the **myip** parameter:

1. Use commas to separate multiple IP addresses in the myip field.

2. If this parameter is not specified, the best IP address the server can determine will be used (some proxy configurations pass the IP in a header, and that is detected by the server). If the IP address passed to the system is not properly formed, it will be ignored and the system’s best guess will be used.

### DIFF
--- a/app/views.py
+++ b/app/views.py
@@ -5,6 +5,7 @@ import traceback
 import re
 import datetime
 import json
+import ipaddress
 from distutils.util import strtobool
 from distutils.version import StrictVersion
 from functools import wraps
@@ -1742,6 +1743,11 @@ def dyndns_update():
     hostname = request.args.get('hostname')
     myip = request.args.get('myip')
 
+    if not hostname:
+        history = History(msg="DynDNS update: missing hostname parameter", created_by=current_user.username)
+        history.add()
+        return render_template('dyndns.html', response='nohost'), 200
+
     try:
         # get all domains owned by the current user
         domains = User(id=current_user.id).get_domain()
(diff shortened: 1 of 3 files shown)


app/views.py
```diff
@@ -1765,37 +1771,61 @@ def dyndns_update():
...
+                if result['status'] == 'ok':
+                    history = History(msg='DynDNS update: updated {0} record {1} in zone {2}, it changed from {3} to {4}'.format(rtype,hostname,domain.name,oldip,str(ip)), detail=str(result), created_by=current_user.username)
+                    history.add()
+                    response='good'
+                else:
+                    response='911'
+                    break
+        elif r.is_allowed_edit():
+            ondemand_creation = DomainSetting.query.filter(DomainSetting.domain == domain).filter(DomainSetting.setting == 'create_via_dyndns').first()
+            if (ondemand_creation != None) and (strtobool(ondemand_creation.value) == True):
```
joniw: This leads to an exception:
```
[ERROR] Exception on /nic/update [GET]
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python3.6/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python3.6/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python3.6/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/usr/lib/python3.6/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python3.6/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "./app/views.py", line 94, in decorated_function
    return f(*args, **kwargs)
  File "./app/views.py", line 1817, in dyndns_update
    if (ondemand_creation != None) and (strtobool(ondemand_creation.value) == True):
  File "./app/models.py", line 734, in __eq__
    return self.setting == other.setting
AttributeError: 'NoneType' object has no attribute 'setting'
```
This is triggered because before the resource type was 'None'. When init the resource type with ipv4 or ipv6, the equality [operator](https://github.com/ngoduykhanh/PowerDNS-Admin/blob/master/app/models.py#L734) triggers an exception as shown above. 

To prevent this you can check with `is not`:

```suggestion
            if (ondemand_creation is not None) and (strtobool(ondemand_creation.value) == True):
```

I will also add a pull request to fix this in the DomainSettings class

ngoduykhanh: Merged this PR to solve the issue.

app/views.py
```diff
@@ -1765,37 +1771,61 @@ def dyndns_update():
...
-    r = Record()
-    r.name = hostname
-    # check if the user requested record exists within this domain
-    if r.exists(domain.name) and r.is_allowed_edit():
-        if r.data == myip:
-            # record content did not change, return 'nochg'
-            history = History(msg="DynDNS update: attempted update of {0} but record did not change".format(hostname), created_by=current_user.username)
-            history.add()
-            return render_template('dyndns.html', response='nochg'), 200
+    def validate_address(address):
```
ngoduykhanh: Can we move this function to another place? `lib/utils.py` for example.

mtdcr:
I've incorporated both suggestions and rebased my branch onto today's master.

ngoduykhanh:
LGTM. Merged your PR. Thanks for the contribution.

PR #458, opened 2019-02-11
fix equality check for DomainSetting when "other" is None

joniw:
When checking a DomainSetting against None with "==", an exception is thrown. See [here](https://github.com/ngoduykhanh/PowerDNS-Admin/pull/455/files#r255691535) for the explanation.


ngoduykhanh:
LGTM. Merging the PR. Thanks.

PR #469, opened 2019-02-23
Improve SAML support

timja:
Hi

I tried to integrate this with AzureAD and hit a few issues, 
I've made some changes to make this flow work better, let me know what you think

- Make SAML_WANT_MESSAGE_SIGNED configurable, AzureAD signs the assertion but wouldn't sign the message
- Add support for a name attribute, i.e. 'Tim Jacomb' using `SAML_ATTRIBUTE_NAME`, which will be mapped into the given and surname fields, AzureAD only has displayname
- Add support for group based admin `SAML_ATTRIBUTE_GROUP` and `SAML_GROUP_ADMIN_NAME`
- Add support for group based accounts `SAML_GROUP_TO_ACCOUNT_MAPPING`
- Don't fail if cert and key aren't present

I've updated the config_template.py

Didn't see anywhere where there was tests for this but let me know if you need me to update anywhere

### DIFF
--- a/app/views.py
+++ b/app/views.py
@@ -293,27 +293,42 @@ def saml_authorized():
         email_attribute_name = app.config.get('SAML_ATTRIBUTE_EMAIL', 'email')
         givenname_attribute_name = app.config.get('SAML_ATTRIBUTE_GIVENNAME', 'givenname')
         surname_attribute_name = app.config.get('SAML_ATTRIBUTE_SURNAME', 'surname')
+        name_attribute_name = app.config.get('SAML_ATTRIBUTE_NAME', None)
         account_attribute_name = app.config.get('SAML_ATTRIBUTE_ACCOUNT', None)
         admin_attribute_name = app.config.get('SAML_ATTRIBUTE_ADMIN', None)
+        group_attribute_name = app.config.get('SAML_ATTRIBUTE_GROUP', None)
+        admin_group_name = app.config.get('SAML_GROUP_ADMIN_NAME', None)
+        group_to_account_mapping = create_group_to_account_mapping()
+
         if email_attribute_name in session['samlUserdata']:
             user.email = session['samlUserdata'][email_attribute_name][0].lower()
         if givenname_attribute_name in session['samlUserdata']:
             user.firstname = session['samlUserdata'][givenname_attribute_name][0]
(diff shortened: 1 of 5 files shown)


timja:
@ngoduykhanh would you be able to take a look when you have time please?

ngoduykhanh:
Hi @timja ,

Thanks for your contribution. The changes look good to me. Howerver there is few places can be adjusted. Please see my code comments.

app/views.py
```diff
@@ -343,7 +355,37 @@ def saml_authorized():
...
+        account = Account(name=clean_name.lower(), description='', contact='', mail='')
+        account.create_account()
+        history = History(msg='Account {0} created'.format(account.name), created_by='SAML Assertion')
+        history.add()
+    return account
+
+
+def uplift_to_admin(user):
+    admin_role = Role.query.filter_by(name='Administrator').first().id
+    if user.role_id != admin_role:
```
ngoduykhanh: We can use `user.role.name != 'Administrator'` directly

app/views.py
```diff
@@ -324,14 +339,11 @@ def saml_authorized():
...
-                history = History(msg='Promoting {0} to administrator'.format(user.username), created_by='SAML Assertion')
-                history.add()
-          else:
+        if admin_attribute_name and 'true' in session['samlUserdata'].get(admin_attribute_name, []):
+            uplift_to_admin(user)
+        elif admin_group_name in user_groups:
+            uplift_to_admin(user)
+        elif admin_attribute_name or group_attribute_name:
             user_role = Role.query.filter_by(name='User').first().id
             if user.role_id != user_role:
```
ngoduykhanh: We can use `user.role.name != 'User'` directly

config_template.py
```diff
@@ -71,6 +71,12 @@
 ### Example: urn:oid:2.5.4.4
 #SAML_ATTRIBUTE_SURNAME = 'urn:oid:2.5.4.4'
 
+## Split into Given name and Surname
```
ngoduykhanh: Can you also sync the new changes to config files in `configs/`

timja: done

ngoduykhanh:
LGTM, merged the PR. Thanks @timja

PR #471, opened 2019-02-28
Fix typo in user profile template

vipera:
Fixes spelling error on the Authentication tab of the user profile.


ngoduykhanh:
Merged, thanks for caching that.

PR #484, opened 2019-03-18
escape special chars when creating group filter.

genericpenguin:
The LDAP search filter used for group queries needs to be escaped so that group names with special characters will not break the search filter in queries. e.g. a query for a group with a Common Name with parens will return 'Bad search filter' and break authentication.


ngoduykhanh:
Thanks for catching that. Merged the PR.

Issue #364, opened 2018-09-22
LUA Records


canuckbrian:
LUA records aren't able to be added or visible in PowerDNS-Admin.

Currently they do not appear in the interface and need to be manually added/edited from the server console using pdnsutil.

ngoduykhanh:
This record is a specific feature from PNDS. Currently we haven't added it into the list yet. There are things need to be done if we want to support it
- Check pdns version
- Check zone / pdns config that has `enable-lua-record` enabled
- Record helper for LUA might need to be added.

nagytech:
I've enabled lua records for all servers to use mysql backend, but I'm not able to see the 'LUA' coded value in the record type drop down.  I can see the value exists in the source code from #488, but there's no indication as to why it's not showing.

OS: Ubuntu Xenial
PDNS Version: 3.4.0 w/ geoip, mysql, lua modules
PDNS Admin Version: 
- ngoduykhanh/powerdns-admin:latest
- DIGEST:sha256:b921f63582b8418c482ae917637c7558fb833fba6ab63009b2c1b739651fb815

I've found that I can inject the LUA value into the dropdown and a LUA script form pops up which enables me to submit the record.  The submission appears successful, but actually fails.  And, as also reported in #385, the UI still doesn't completely support LUA records and the server page fails to load after the fact.

<img width="1067" alt="image" src="https://user-images.githubusercontent.com/8542033/76444001-692fff00-6417-11ea-8355-a9b596acd56b.png">

<img width="1080" alt="image" src="https://user-images.githubusercontent.com/8542033/76444228-be6c1080-6417-11ea-9b61-78641bfe8e34.png">

I think this feature still needs a bit of work - hopefully it's just a configuration issue.  Maybe someone  knows where it's up to?

nagytech:
Sorry, I just realised you have to go into Settings > Forward Zone and enable the LUA type.

I have no issue viewing zones that have LUA records, this means that #385 can probably be closed now.

PR #488, opened 2019-03-23
Basic LUA record support

olemathias:
Solves part of #364

- Do not see the need to implement a powerdns version/config check as of now. The record type is default disabled. Can be added later.
- Simple record helper is added.
- Little formatting of javascript in domains.html to clean things up a little.


ngoduykhanh:
Hi, Thanks for submitting the PR. Your changes look good to me. Merging it.
LUA has been supported since 4.2, so older version will not work. For the version checking, we can ignore it for now as the record is disabled by default.

PR #490, opened 2019-03-28
Records entries must not make database requests for domain listing page

sayoun:
When you want to display records for a domain, it will make 4 SQL requests for each records entry, to retrieve Settings fields which are not used at all in the domain page. This results in a really slow processing time, in our case for a zone with 4000 records, it will make 16000 queries.

This PR will create a new class just for storing records entries, which is sufficient for the domain page template.

Before this PR, the rendering of our 4000 records took 7 minutes, and now it only takes 4 seconds.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1377,6 +1377,29 @@ class AccountUser(db.Model):
         return '<Account_User {0} {1}>'.format(self.account_id, self.user_id)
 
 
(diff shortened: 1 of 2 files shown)


ngoduykhanh:
LGTM. Thanks for catching that and submit the PR. Merging it.

Issue #479, opened 2019-03-05
applying changes to large domain: "Error when applying new changes"


megabert:
When I try to make changes at a large domain, I get the following error(Red Error Message window):

"Error: Error when applying new changes"

The changes are not applied into the pdns-database. The powerdns-admin interface however shows the new, not applied data.

When I look into /powerdns-admin/logs/log.txt I get the following error:

```2019-03-05 13:25:30,568 - app.views - ERROR - Cannot apply record changes. Error: (_mysql_exceptions.DataError) (1406, "Data too long for column 'detail' at row 1") [SQL: 'INSERT INTO history (msg, detail, created_by, created_on) VALUES (%s, %s, %s, %s)'] [parameters: ('Apply record changes to domain mydomain.de', '{"serial": "2019020521", "record": [{"record_data": "1.2.3.4", "record_name": "*.cluster", "record_status": "Active", "record_type": "A", "reco ... (133402 characters truncated) ...  "86400"}, {"record_data": "1.2.3.5", "record_name": "testshop", "record_status": "Active", "record_type": "A", "record_ttl": "86400"}]}', 'admin', datetime.datetime(2019, 3, 5, 13, 25, 30, 546552))] (Background on this error at: http://sqlalche.me/e/9h9h)```

The Domain holds 1000 records. (Other, smaller Domains are working perfectly).

That is obviously not the only error happening, since it states it can not insert a record into the history table (of powerdns-admin database). But there's nothing about the not accomplished/failed api call for powerdns.

Environment: 
- Docker
- MariaDB 10.3.12
- powerdns-admin version: current version from github

p53:
this seems to be related to this https://github.com/PowerDNS/pdns/issues/5992 chunk encoding could fix it

megabert:
Thanks for your answer p53.

As far as I see #5992 and #1469 are about curl and maximum header size/length. 

This error is a mysql error.

Dolphi2:
I have exactly the same problem with large domains!

The colomn 'detail' in table 'history' is of type text which has a maximum of 65535 charachters. Changing it to LONGTEXT solved the problem for me.
MariaDB [pda]> describe history;
+------------+--------------+------+-----+---------+----------------+
| Field      | Type         | Null | Key | Default | Extra          |
+------------+--------------+------+-----+---------+----------------+
| id         | int(11)      | NO   | PRI | NULL    | auto_increment |
[...]

PR #492, opened 2019-04-04
Trim history detail to max 1000 characters

p53:
#479

### DIFF
--- a/app/blueprints/api.py
+++ b/app/blueprints/api.py
@@ -99,9 +99,10 @@ def api_login_create_zone():
         logging.debug("Request to powerdns API successful")
         data = request.get_json(force=True)
 
+        history_detail = json.dumps(data)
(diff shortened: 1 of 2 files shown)


vampik:
There is a bug. Trimmed history cannot be viewed in web-ui.
Browser console shows errors like:
SyntaxError: JSON.parse: unexpected end of data at line 1 column 1001 of the JSON data
SyntaxError: JSON.parse: unterminated string at line 1 column 1001 of the JSON data
SyntaxError: JSON.parse: end of data while reading object contents at line 1 column 1001 of the JSON data
etc.
It seems that web-ui expects a valid JSON, but trimming to 1000 characters breaks JSON format.

ngoduykhanh:
History data is now reduced with #604. Closing.

PR #500, opened 2019-04-29
Added endpoint synchronize domains

in0th3p:
added API-Endpoint and swagger definition for /sync_domains to update/synchronize domains in background


ngoduykhanh:
LGTM, Merging your PR. Thanks.

PR #501, opened 2019-04-29
added TimeStamp to logging

in0th3p:
Timestamp was missing when developing with Docker-compose so we added it.


ngoduykhanh:
Merging. Thanks.

PR #531, opened 2019-06-29
Standard compliant HTTP Response codes

pfak:
Hello,

Presently the API returns non-standard complaint HTTP response codes. This causes reverse proxies such as Apache 2.4 to drop the responses:

[Thu Jun 27 18:23:12.684170 2019] [proxy_http:error] [pid 11733] (70014)End of file found: [client 216.19.183.221:53670] AH01102: error reading status line from remote server 127.0.0.1:8000
[Thu Jun 27 18:23:12.684611 2019] [proxy:error] [pid 11733] [client x.x.x.x:53670] AH00898: Error reading from remote server returned by /powerdns-admin/api/v1/pdnsadmin/apikeys

I'm proposing the following changes to the HTTP response codes, as I cannot see them being used anywhere else I don't think this should break any workflows.

### DIFF
--- a/app/errors.py
+++ b/app/errors.py
@@ -19,7 +19,7 @@ class StructuredException(Exception):
 
 
 class DomainNotExists(StructuredException):
-    status_code = 1000
+    status_code = 404
 
     def __init__(self, name=None, message="Domain does not exist"):
         StructuredException.__init__(self)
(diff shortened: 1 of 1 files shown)


Armadill0:
The correct code a forbidden response would be 403 instead of 401 I think.

ngoduykhanh:
Thanks for submitting the PR.
Agree with the above commend, the code for a forbidden should be 403. Can we change it?

pfak:
I made the change and updated the commit.

ngoduykhanh:
Merged it. Thanks for your contribution!

PR #535, opened 2019-07-12
a Dockerfile for production run

Jamesits:
The current installation method is rather dirty and involves a lot of 3rd party tools. I wrote a Dockerfile for a clean installation of PowerDNS-Admin so it can be used in production with a clean environment. 

Usage for reference:

Create a env file `pdnsadmin-env`:
```
SECRET_KEY=your-very-secret-key
TIMEOUT=10
LOG_LEVEL=DEBUG
LOG_FILE=logfile.log
SALT=$2b$12$somerandomsalt
SQLALCHEMY_TRACK_MODIFICATIONS=True
SQLALCHEMY_DATABASE_URI=postgres://user:password@your.server.address:5432/database-name
```

Then run the prebuilt docker image:

```shell
docker run -p 9191:80 --env-file pdnsadmin-env jamesits/powerdns-admin:latest
```

Visit `http://localhost:9191` and you are ready to go.

### DIFF
--- a/docker/Production/Dockerfile
+++ b/docker/Production/Dockerfile
@@ -4,7 +4,7 @@ LABEL maintainer="k@ndk.name"
 ENV LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 LANGUAGE=en_US.UTF-8
 
 RUN apt-get update -y \
-	&& apt-get install -y --no-install-recommends apt-transport-https locales locales-all python3-pip python3-setuptools python3-dev curl libsasl2-dev libldap2-dev libssl-dev libxml2-dev libxslt1-dev libxmlsec1-dev libffi-dev pkg-config build-essential libmariadb-dev-compat \
+	&& apt-get install -y --no-install-recommends apt-transport-https locales locales-all python3-pip python3-setuptools python3-dev curl libsasl2-dev libldap2-dev libssl-dev libxml2-dev libxslt1-dev libxmlsec1-dev libffi-dev build-essential libmariadb-dev-compat \
     && curl -sL https://deb.nodesource.com/setup_10.x | bash - \
     && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
     && echo "deb https://dl.yarnpkg.com/debian/ stable main" > /etc/apt/sources.list.d/yarn.list \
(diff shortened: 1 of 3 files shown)


ngoduykhanh:
LGTM. Merged your PR. Thanks for your contribution.

Jamesits:
@ngoduykhanh Could you please set up an official docker automated build on hub.docker.com? This will make it more convenient for us. Please set Dockerfile location to `docker/Production/Dockerfile` and leave other as default. 

image

ngoduykhanh:
Yep, I have just created a repo here: https://hub.docker.com/r/ngoduykhanh/powerdns-admin and do the first build.
However, the image is quite huge, I am gonna adjust the Dockerfile to minimize the size and update the README

Jamesits:
The compilers take a lot of places, maybe we can purge them after `pip install`. But this does not affect image size unless you squash them into one layer, which is not good for build caching. 

Another way to achieve this is multistage build -- but I'm not sure how to get those python native modules out of a container.

Tobias-Fu:
Is there a reason why docker repo ngoduykhanh/powerdns-admin does not include support for posgreaql DB? Only MySQL, MariaDB and sqlite is suported. The repo ixpict/powerdns-admin-pgsql referenced on [WIKI](https://github.com/ngoduykhanh/PowerDNS-Admin/wiki/Using-PowerDNS-Admin-with-PostgreSQL) seems to be not getting maintained anymore. 

I would appreciate if posgresql support could get included in the official docker file.

ngoduykhanh:
@Tobias-Fu The [Dockerfile](https://github.com/ngoduykhanh/PowerDNS-Admin/blob/master/docker/Dockerfile#L79) release image already had `py3-psycopg2` package. It should work, no?

Tobias-Fu:
OK. Thank you for confirmation that it is in and that it should work.

It did not work for me under CentOS 8 with podman. When I try to start the container with `-e SQLALCHEMY_DATABASE_URI=postgresql://...` it stops to status "exited". But with `-v pda-data:/data` instead it is running. 

It was only my assumption that postgresql support is not in because the WIKI refers to a different docker image. Now I can continue searching for the reason on my system. Connection I want to create to posgresql running directly on CentOS, not in a container.

Tobias-Fu:
Solved.
Container ngoduykhanh/powerdns-admin is running with podman under Centos 8 and is connecting to postgresql.
The problem was a connection error to the db. I did not expect that the container will crash in such a case.

PR #547, opened 2019-07-24
fix #542

### LINKED ISSUE #542: Hardcoded IP address in production docker
When running image with network mode host - it is using 0.0.0.0:80
https://github.com/ngoduykhanh/PowerDNS-Admin/blob/master/docker/Production/entrypoint.sh#L13
I can't change it

Jamesits:


ngoduykhanh:
LGTM. Thanks for the contribution

PR #565, opened 2019-10-11
Sort accounts pull down menu in domain_add.html

jntdd:


bryanpedini:
I think #766 is a better way of handling this type of changes, since it pulls sorted entries directly from the database, whenever and everywhere they are needed.

zoeller-freinet:
The already merged #766 does query the accounts ordered by name from the database, but only if the current user either has the Administrator or the Operator role. #1049 proposes to also order them by name if the current user has the User role. That should cover all relevant cases apart from the `/pdnsadmin/accounts` API endpoint.

PR #574, opened 2019-10-23
Update Dockerfile PowerDNS-Admin

Revellski:
Workaround in upstream image (ubuntu:16.04) to avoid Hash Sum mismatch errors during build.


ngoduykhanh:
PR #586 updated the Dockerfile also.

PR #577, opened 2019-11-05
forward get zones to real powerdns api

paepke:
There is already an endpoint which is correctly prefixed by /pdnsadmin so this one (/servers/<string:server_id>/zones) should correctly be forwarded to the real powerdns api.

### DIFF
--- a/app/blueprints/api.py
+++ b/app/blueprints/api.py
@@ -504,8 +504,15 @@ def api_create_zone(server_id):
 @api_blueprint.route('/servers/<string:server_id>/zones', methods=['GET'])
 @apikey_auth
 def api_get_zones(server_id):
-    resp = helper.forward_request()
-    return resp.content, resp.status_code, resp.headers.items()
+    if server_id == 'powerdns':
+        if g.apikey.role.name not in ['Administrator', 'Operator']:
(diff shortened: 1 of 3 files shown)


paepke:
just saw the refactoring pull request #586 - its still applicable there

ngoduykhanh:
Yes, I am working on the refactoring. I included your changes [here](https://github.com/ngoduykhanh/PowerDNS-Admin/blob/code_refactor/powerdnsadmin/routes/api.py#L502-L513). I think it is better with naming "powerdns-admin" instead of "powerdns" when we do the query to PowerDNS-Admin directly, isn't it?

paepke:
Yes, powerdns-admin is better.
As mentioned above

> There is already an endpoint which is correctly prefixed by /pdnsadmin so this one (/servers/string:server_id/zones) should correctly be forwarded to the real powerdns api.

we should not use the same api endpoint for returning different data structures. If there is no known logic that depends on requesting powerdns-admin zone data via "/servers/string:server_id/zones"-endpoint the we should completely forward every request to the real powerdns api and get rid of the condition I've introduced.

ngoduykhanh:
Fixed in #586

PR #582, opened 2019-11-18
Add Systemd service example

davispuh:


sshipway:
I think the systemd example in this PR no longer works with the new version of pDNS-a, as the path name has changed.  Also, it would benefit from more dependency definitions

We have this:
```
[Unit]
Description=PowerDNS web administration service
Requires=powerdns-admin.socket
Wants=network.target
After=network.target mysqld.service postgresql.service slapd.service mariadb.service

[Service]
PIDFile=/run/powerdns-admin/pid
User=pdnsa
Group=pdnsa
WorkingDirectory=/opt/powerdns-admin
ExecStart=/usr/bin/gunicorn-3.6 --workers 4 --log-level info --pid /run/powerdns-admin/pid --bind unix:/run/powerdns-admin/socket "powerdnsadmin:create_app(config='config.py')"
ExecReload=/bin/kill -s HUP $MAINPID
ExecStop=/bin/kill -s TERM $MAINPID
PrivateTmp=true
Restart=on-failure
RestartSec=10
StartLimitInterval=0

[Install]
WantedBy=multi-user.target
```

Then there is the powerdns-admin.socket and tmpfiles.d as in the supplied documentation

davispuh:
I created this PR before I had seen wiki so I it probably can just be closed as there's already example [Running PowerDNS Admin with Systemd, Gunicorn and Nginx](/ngoduykhanh/PowerDNS-Admin/wiki/Running-PowerDNS-Admin-with-Systemd,-Gunicorn--and--Nginx)

Also it's just example, different distributions might install it in different locations.

zoeller-freinet:
Yes, please resort to the wiki for systemd unit examples.

PR #591, opened 2019-12-06
Add explicit Azure oauth support

sshipway:
The generic oauth does not work with Azure, since Microsoft have a differently named API function and different attributes returned.  This PR adds support for this and some instructions on how to link them.
Note that Azure oAuth will only work if you have HTTPS, and since this is often done via an nginx wrapper the url_for has to be passed the _scheme='https' option.
Tested here successfully.


ngoduykhanh:
Thanks for your contribution. Since the code was refactored. I make another commit to adjust your PR and have some small fixes.

PR #600, opened 2019-12-11
add history entries for rrsets

paepke:
This pull request adds history entries to rrset updates and deletes via the api to the history table of PDA.


ngoduykhanh:
You are creating history for each record in a `for` loop. I think it is not necessary because the "detail" already shows the whole  rrset data.

PR #605, opened 2019-12-14
Most of SAML integration fixed

Neven1986:
- Variable references inside SAML class were fixed
    - Function signatures inside SAML class were fixed
    - Redirect URL for /saml/login path was modified (saml_authorized -> index.saml_authorized)

    Current status is that SAML metadata can be generated under /saml/metadata and communication to SAML iDP is working

    Problems remaining:
        - SAML Response doesn't contain any attributes (There is no AttributeStatement on the Response). Probably it is problem on iDP side. I'll need to take other SAML iDP.
        - Background thread in retrieve_idp_data() cannot be spawned, this part is currently commented out, old code needs to be revisited. Will to try to resolve it soon.

### DIFF
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -527,7 +527,7 @@ def saml_login():
     req = saml.prepare_flask_request(request)
     auth = saml.init_saml_auth(req)
     redirect_url = OneLogin_Saml2_Utils.get_self_url(req) + url_for(
-        'saml_authorized')
+        'index.saml_authorized')
     return redirect(auth.login(return_to=redirect_url))
 
 
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -663,7 +663,7 @@ def saml_authorized():
         user.update_profile()
         session['authentication_type'] = 'SAML'
         login_user(user, remember=False)
-        return redirect(url_for('index'))
+        return redirect(url_for('index.login'))
     else:
         return render_template('errors/SAML.html', errors=errors)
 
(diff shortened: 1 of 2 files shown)


Neven1986:
SAML seems to be reintegrated now. I conducted few tests with samlidp.io as iDP.
Please review this PR and if you find it OK merge it.

There are few things which should be implemented, but as new features and as separate pull requests. I will try to take care of it as well in coming week or two.

ngoduykhanh:
Hi @Neven1986 , thanks for submitting the PR. The code looks good to me. I wanted to test with samlidp.io as well. Do you know which SAML_SP_ENTITY_ID value and certificate I should use for it?

Neven1986:
Hi @ngoduykhanh,

SAML SP ENTITY ID is pretty arbitrary thing and can be almost any URI derived from your domain or your domain itself. In my case that was http://pda.<my.domain>:9191/ - simple as that...

You can use any self-signed x509 certificate... I used my own generated one and I used in second test certificates generated by PDA from certutils.py

Just as side note... With samlidp.io there is one peculiar thing. In order to get attributes in response you need to explicitly request them by defining AttributeConsumingService attributes in your metadata... Which is currently empty 'sp settings' substructure in saml.py. For testing purposes I needed to manually define this in the code since it's not configurable via default_config.py. This is one of improvements which I talked about, that I would like to implement.

I did additional test with my company's iDP and it passed went through as well. The only difference is that this iDP will return all attributes without requesting them specifically in SP metadata. A lot of things are dependent on iDP config itself. 

If you need more heads-up on this I can send you my example metadata.

ngoduykhanh:
Hi @Neven1986 , thanks for the explanation and tests.
I wanted to keep the default_config file minimal. If you look at the configs/development.py file, you would see more config related to SAML.

I am going to merge your PR. If you have any adjustments in the feature, feel free to create another one.

PR #607, opened 2019-12-15
Update dockerfile to reduce image size

chrisss404:
* use alpine as base image
* employ multi stage build to get rid of build artefacts


paescuj:
Looks good, but there is no need to cleanup build dependencies in the build stage.

chrisss404:
> Looks good, but there is no need to cleanup build dependencies in the build stage.

Yep, you are right. However, it doesn't hurt?

paescuj:
No, it doesn't. Just takes a second longer :smiley:

PR #612, opened 2019-12-19
SAML certificate fix and enhancement

Neven1986:
Problems resolved:

- Method create_self_signed_cert() was invoked nowhere. This puts parameter "SAML_SIGN_REQUEST" description in configs/development.py as incorrect
- Method create_self_signed_cert() was returning error while trying to write out certificate and private key. File handler was opened for writing out TEXT instead of BINARY data

Enhancements:

- Two new parameters are introduced SAML_CERT_FILE and SAML_KEY_FILE. User can now explicitly define own certificate and key file anywhere on file-system.
- If parameters mentioned in previous bullet aren't explicitly defined, in PowerDNS-Admin root directory self-signed certificate will be created.
- Certificates will be used or generated in any case, because in saml.py there are explicit parameters defined which require certificate/key in order to work normally. If they aren't, exception will be thrown. Examples of parameters defined in saml.py requiring certificate: wantAssertionsEncrypted, signMetadata, wantAssertionsSigned.

If you have any questions don't hesitate to ask...

### DIFF
--- a/configs/development.py
+++ b/configs/development.py
@@ -82,8 +82,20 @@ SAML_ENABLED = False
 # SAML_SP_ENTITY_ID = 'http://<SAML SP Entity ID>'
 # SAML_SP_CONTACT_NAME = '<contact name>'
 # SAML_SP_CONTACT_MAIL = '<contact mail>'
-# #Cofigures if SAML tokens should be encrypted.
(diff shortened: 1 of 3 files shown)


ngoduykhanh:
LGTM. Merging the PR. Thanks for your contribution.

PR #642, opened 2020-01-25
Add dashboard search placeholder to describe search patterns

eht16:
Since ^ and $ for searching from start and to end is supported, let's
tell the user about it.


ngoduykhanh:
LGTM. Merged the PR. Thanks

PR #643, opened 2020-01-25
Add missing 'logging' imports

eht16:
Those imports were missing and lead to errors in case of other errors which should be logged but `logging` was not imported yet.


ngoduykhanh:
Thanks for rasing the issue. Actually I missed that "logging" module in more than one place during the code refactoring... I've made another PR for this - #652.

eht16:
Yeah, even better this way. Thanks.

PR #644, opened 2020-01-25
Add new setting to verify outgoing SSL connections

eht16:
The new setting 'verify_ssl_connections' tells the requests library to
verify secured outgoing HTTP connections.
Usually verifying is desired and helps to reveal configuration
problems. It also disables an ugly warning when HTTPS connections
are made without verification.

Most people probably connect to the PowerDNS API via plain HTTP connections anyway. But for those few who connect via HTTPS, SSL verification should be enabled unless they explicitly disable it.


ngoduykhanh:
LGTM. Merging your PR. Thanks for your contribution.

PR #645, opened 2020-01-25
Add log message after remote server request

eht16:
This might be handy to detect remote server problems or debugging
timeouts and the like.

### DIFF
--- a/powerdnsadmin/lib/utils.py
+++ b/powerdnsadmin/lib/utils.py
@@ -1,3 +1,4 @@
+import logging
 import re
 import json
 import requests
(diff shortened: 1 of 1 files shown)


ngoduykhanh:
Ok. But I think the log level should be DEBUG, no?

eht16:
Yes, oops :).
I changed the log level to debug.

Regarding #652: in theory we use `current_app.logger` as well instead of the `logging` module.
But I assumed the `utils` should be kept independent of Flask or other components by design and so I went the way using the plain `logging` module. Not sure where is your preference.

ngoduykhanh:
Yes, I am OK with using the plain `logging` module in the independent libs. Merging now, thanks.

PR #648, opened 2020-01-26
Introduce PDNS_ADMIN_LOG_LEVEL to change the log level

eht16:
By setting the environment variable PDNS_ADMIN_LOG_LEVEL to a Python
support log level, PowerDNS-Admin will use that log level.

If the environment variable is missing, `WARNING` will be used what is the current default.

I like to set the log level to info (which was the default in earlier versions) and with this change this is possible again.


ngoduykhanh:
LGTM. Merging the PR. Thanks.

PR #655, opened 2020-02-02
Perform domain sync within a single transaction

### LINKED ISSUE #428: When domain is deleted outside of PowerDNS-Admin the worker timeouts on dashboard
When a PowerDNS domain is deleted outside of PowerDNS-Admin, i.e. in the PowerDNS back-end directly, the PowerDNS-Admin gunicorn worker times out (already tried raising the limit) on the /dashboard page:

```
Dec 19 13:49:49 dns gunicorn: [DEBUG] Update domains in foreground
Dec 19 13:49:49 dns gunicorn: [DEBUG] Starting new HTTP connection (1): 127.0.0.1:8081
Dec 19 13:49:49 dns gunicorn: [DEBUG] http://127.0.0.1:8081 "GET /api/v1/servers/localhost/zones HTTP/1.1" 200 2
Dec 19 13:51:20 dns gunicorn: [2018-12-19 13:51:20 +0100] [18743] [CRITICAL] WORKER TIMEOUT (pid:18829)
Dec 19 13:51:20 dns gunicorn: [2018-12-19 13:51:20 +0100] [18829] [INFO] Worker exiting (pid: 18829)
Dec 19 13:51:20 dns gunicorn: [2018-12-19 13:51:20 +0100] [21424] [INFO] Booting worker with pid: 21424
```

Removing the domains also from the `domains` table in the PowerDNS-Admin MySQL database resolves the issue. 

It seems that this is a bug in PowerDNS-Admin which renders PowerDNS-Admin unusable if a domain is deleted on the PowerDNS nameserver.

eht16:
This increases data inconsistency and also performance of the whole
update process: SqlAlchemy can use its own caching better with a single
transaction as it does not need to mark already fetched objects as dirty
and so re-fetch them again on later use. So after deleting a domain,
we can still use the previously fetched domain objects for the update
operations afterwards. Otherwise SqlAlchemy had to fetch each domain
object again with single SELECT statements which leads to bad
performance with many domains (>> 1000).

The introduced "do_commit" variable controls whether the database
changes are to be committed by the caller or in the called method
itself. So we can use a single transaction for the Domain.update()
method while still preserving the previous behavior for other callers.

Closes #428.

### DIFF
--- a/powerdnsadmin/models/domain.py
+++ b/powerdnsadmin/models/domain.py
@@ -138,7 +138,7 @@ class Domain(db.Model):
                 should_removed_db_domain = list(
                     set(list_db_domain).difference(list_jdomain))
                 for domain_name in should_removed_db_domain:
-                    self.delete_domain_from_pdnsadmin(domain_name)
+                    self.delete_domain_from_pdnsadmin(domain_name, do_commit=False)
             except Exception as e:
                 current_app.logger.error(
                     'Can not delete domain from DB. DETAIL: {0}'.format(e))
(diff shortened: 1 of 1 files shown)


ngoduykhanh:
Hi @eht16 , your solution looks good to me. I didn't think about that. Merging it now. Thanks for your contribution.

Issue #513, opened 2019-05-21
posixGroup support for LDAP_GROUP_SECURITY


Armadill0:
I just checked out PR #512 which allows LDAP_GROUP_SECURITY with groupOfNames.

My problem is, that we only use posixGroups which differ in using only the username instead of the full FDN of the LDAP user.

I now want to suggest to add a drop down element on the LDAP settings page where you can choose between "posixGroup" and "groupOfNames" layout to use for the group security feature.

The code itself should be pretty straight forward because the only difference is the usage of self.username instead of ldap_username in the groupSearchFilter. This could be easy achieved by an intersection which uses the value from the formerly introduced group layout dropdown field.

I'm not a great developer myself, but I think I could provide the code for this after #512 has been merged.

Any other suggestions or ideas?

johnwarburton:
Have you tried putting this as the Group Filter? I've never used it but I think this should work but I do not have posixgroups to check
```(objectClass=posixGroup)```

Armadill0:
I tried that, but it doesn't work because the username which is being used is in the FDN format (uid=<username>,ou=people,cn=example,cn=org) and the posixGroup format needs <username> only.

tingvold:
Any chance #512 could get merged? Just hit this issue today.

jbe-dw:
As said in the linked PR you're not using the software anymore. Closing.

PR #661, opened 2020-02-09
Implement posixGroup support for LDAP auth

Armadill0:
As described in #513 this implements support for the LDAP group member matching by posixGroup instead of the groupOfNames membership identification.

Screenshot of the filters section:
PDA_ldap-posixGroup-support

### DIFF
--- a/powerdnsadmin/models/setting.py
+++ b/powerdnsadmin/models/setting.py
@@ -44,7 +44,6 @@ class Setting(db.Model):
         'ldap_admin_username': '',
         'ldap_admin_password': '',
         'ldap_filter_basic': '',
-        'ldap_filter_group': '',
         'ldap_filter_username': '',
         'ldap_filter_groupname': '',
         'ldap_group_matching': 'groupOfNames',
--- a/powerdnsadmin/models/setting.py
+++ b/powerdnsadmin/models/setting.py
@@ -47,6 +47,7 @@ class Setting(db.Model):
         'ldap_filter_group': '',
         'ldap_filter_username': '',
         'ldap_filter_groupname': '',
+        'ldap_group_matching': 'groupOfNames',
         'ldap_sg_enabled': False,
         'ldap_admin_group': '',
         'ldap_operator_group': '',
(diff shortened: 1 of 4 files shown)


ngoduykhanh:
Hi @Armadill0 , Thanks for your contribution. In general, your idea looks good to me. I just have a few comments:

- Regarding the UI part, since the Group related configs are only used for GROUP SECURITY feature, should we rearrange it by moving those setting to GROUP SECURITY section?
- Do we still need **Group filter** (`ldap_filter_group`) when `ldap_group_matching` is available? I think its code can be generated from the user's ldap_group_matching selection. My idea is to keep the setting simple as much as possible so it won't confuse the user.

Armadill0:
Hi @ngoduykhanh,

thanks for your message. Your suggestion to rearrange all group settings into the group security section sounds reasonable. I'll do this tomorrow.

I'm not sure if we still need the group filter. My intention was to integrate the feature without breaking existing installations where someone maybe used a special setup that does not fit just normal posixGroups or groupOfNames. I'm unsure if this really is an issue.

Armadill0:
@ngoduykhanh Sorry for the delay, I found no time to test the changes.

I completely replaced ldap_filter_group. Should we also ensure the old configuration value is going to be removed to avoid conflicts with future usage of such a field? If yes, how can I achieve this?

righter83:
Any news on that? seems I have a similar problem with that.

jbe-dw:
Hello,

I would not remove the group filter as it may be useful (You may want to filter out collected groups from a LDAP branch based on another attribute than objectClass).

If you have a selection box, you may want to process the filter inserted in the filter field to replace a potential objectClass filter with the right one instead. 

This would let the user choose between posix or gon safely because it will not process an "objectClass=groupOfNames"  filter if the user chose posix and vice-versa. This would require a bit of extra code to explode the filter and replace the objectClass filter properly.

thomas-amler-idnow:
I'm not using PowerDNSAdmin anymore. Can be closed or reused by someone else.

Armadill0:
I'm not using PowerDNSAdmin anymore. Feel free to take the changes from my branch and implement it in a better way.

PR #664, opened 2020-02-11
Bug: Creating multiple records with single comment

barnarj:
We encountered an issue when creating multiple records of the same name with different data.  The code block which executes to create multiple RecordEntry objects was attempting to use the incremented index value despite the fact that there is only ever a single comment associated with the records.

Hard-coding this to the first element corrected the failure and was an effective fix for our use cases.

### DIFF
--- a/powerdnsadmin/routes/domain.py
+++ b/powerdnsadmin/routes/domain.py
@@ -105,7 +105,7 @@ def domain(domain_name):
                         status='Disabled' if record['disabled'] else 'Active',
                         ttl=r['ttl'],
                         data=record['content'],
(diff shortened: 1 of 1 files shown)


barnarj:
I know this fix is a dirty hack; the underlying issue is moreso that we can send a list of create requests that link to a single comment.

Depending on your preference, I see a couple ways of mitigating the underlying issue:

1.  Analyze the length of the comment list; if 1, then we handle creating with a single comment, else if length of comment list equals the length of the create requests, process iteratively.

2.  Update the API's strictness in accepting comments

sshipway:
See PR #739 which I believe fixes this problem as well

Issue #666, opened 2020-02-26
Response "Query q can't be blank" when trying to use a search API


pprietosanchez:
Hello,

I get the error when I try to use the search API, the command I use is like this:

 `curl -H 'X-API-KEY: THEKEY' -X GET 'http://HOST/api/v1/servers/localhost/search-data?q=somehost&max=1&object_type=record'`

and the response is '{"error": "Query q can't be blank"}'

Can you help me, please?

Thank You!

Regards!

pprietosanchez:
I've sniff traffic between client<->pda and pda<->pdns and I can see that the request made to the pdns doesn't have the parameters passed in the original request (client<->pda)

The traffic between client<->pda is:
`GET /api/v1/servers/localhost/search-data?q=XXXXXXX HTTP/1.1`

Ans the traffic between pda<->pdns is:
`GET /api/v1/servers/localhost/search-data HTTP/1.1`

If I do the request to the pdns it works well.

Regards

pprietosanchez:
Found the issue:

https://github.com/ngoduykhanh/PowerDNS-Admin/pull/668

PR #668, opened 2020-02-28
Correct request method

pprietosanchez:
I recently opened this issue: https://github.com/ngoduykhanh/PowerDNS-Admin/issues/666

I've found the issue, the method of flask "request" in order to call de pdns api with the parameter is request.full_path rather than request.path in the file helper.py.

I hope this pull request be useful.

Regards.


ngoduykhanh:
Nice catch! Thanks for submitting the fix. Merging your PR now...

PR #671, opened 2020-03-03
Add missing format arguments for error message

eht16:


eht16:
I assume the format arguments were just forgotten to be added.

ngoduykhanh:
Yes, you are right. Thanks for submitting the fix

PR #675, opened 2020-03-05
Fix token name in session for OIDC and GitHub authentication

eht16:
Seems like typos during code refactoring.
I didn't test the changes but I think now it's better than before with setting wrong token names in the session.


ngoduykhanh:
Oh yes, it was a mistake during code refactoring. Thank you for catching it. Merging now...

Issue #676, opened 2020-03-06
Comments are not supported on all pdns backends


sahoahfoa:
I am unable to add new records with the v0.2.1 release since the [LMDB backend for pdns](https://doc.powerdns.com/authoritative/backends/lmdb.html) does not support comments. 

This is the error that appears in the log:
`[2020-03-06 07:19:48,490] [record.py:311] ERROR - Cannot apply record changes with adding rrsets step. PDNS error: Hosting backend does not support editing comments.`

For the time being I've edited models/record.py to remove the comments field from new_rrsets / del_rrsets in the apply method.

ngoduykhanh:
The PR got merged. Closing the issue.

PR #679, opened 2020-03-07
Remove blank comments from rrsets before sending for compatibility

sahoahfoa:
As per #676 when submitting changes to a backends that doesn't support comments the api call will fail. This is really only an issue with the [LMDB backend](https://doc.powerdns.com/authoritative/backends/lmdb.html) as most of the other traditional backends support comments. 

I propose removing the comments field from the api request if the comment is blank. This should allow the request to succeed if the backend supports it and gives a clear error if it fails. 
image

I refrained from creating a setting to remove comments as it would silently delete them without notifying the user. Full disclosure I haven't tested this on other backends. This was a quick fix, so suggestions are welcome.

### DIFF
--- a/powerdnsadmin/models/record.py
+++ b/powerdnsadmin/models/record.py
@@ -272,6 +272,15 @@ class Record(object):
         # Get the list of rrsets to be added and deleted
         new_rrsets, del_rrsets = self.compare(domain_name, submitted_records)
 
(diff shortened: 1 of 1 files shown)


ngoduykhanh:
Merged. Thanks for submitting the PR.

PR #681, opened 2020-03-09
Remove $SCRIPT_ROOT from 'admin.create_template_from_zone' URL

eht16:
As the URL here is contructed from Flask's "url_for" which already
takes the script root into account, we do not need to add it here
explicitly. This would result in a duplicate script root otherwise.


ngoduykhanh:
You're right. Merging your PR. Thanks

PR #685, opened 2020-03-12
Update Alpine 3.10 -> 3.11, update pip, add tzdata

christianbur:
In this PR I updated the Dockerfile from Alpine 3.10 to 3.11. Additionally pip was updated to avoid the error message of the outdated pip version. To set the correct local timezone (environment variable "TZ" must be set) tzdata was added. 

Update:
Unfortunately I just found this bug.
https://github.com/benoitc/gunicorn/issues/2091


christianbur:
oh, 

https://github.com/benoitc/gunicorn/issues/2091

PR #686, opened 2020-03-12
Update Alpine 3.10 -> 3.11; update pip, tzdata, gunicorn

christianbur:
In this PR I updated the Dockerfile from Alpine 3.10 to 3.11. Additionally pip was updated to avoid the error message of the outdated pip version. To set the correct local timezone (environment variable "TZ" must be set) tzdata was added.
Due to the bug (https://github.com/benoitc/gunicorn/issues/2091) I also had to update gunicorn. 

Please test before merging.


ngoduykhanh:
I tested it. LGTM. Merging your PR.  Thanks for your contribution.

PR #690, opened 2020-03-25
SAML debug attributes

Atisom:


powerdnsadmin/services/saml.py
```diff
@@ -161,7 +161,7 @@ def init_saml_auth(self, req):
         settings['security']['requestedAuthnContext'] = True
         settings['security'][
             'signatureAlgorithm'] = 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256'
-        settings['security']['wantAssertionsEncrypted'] = True
+        settings['security']['wantAssertionsEncrypted'] = False
```
ngoduykhanh: Shouldn't it be `settings['security']['wantAssertionsEncrypted'] = current_app.config.get(
            'SAML_ASSERTION_ENCRYPTED', True)` ?

Atisom: Thank you for your response!
I don't want to add this commit to the PR, only the previously (https://github.com/ngoduykhanh/PowerDNS-Admin/pull/690/commits/a581aa3cf230d03908480a70519b63ccebbe5363)

Unfortunately I am not able to pass the "False" value via docker compose to the app, the app does not get it. So it just a quick workaround for me. Or do you have any idea?

```
# docker exec 9c99d2265658 printenv | grep SAML_ASSERTION_ENCRYPTED
SAML_ASSERTION_ENCRYPTED=False
```

ngoduykhanh: Hmm, it should work. What value of SAML_ASSERTION_ENCRYPTED did you get in the container's file `./powerdnsadmin/docker_config.py`

Atisom: for debugging, I insert a debug line to the `routes/index.py`

```
@index_bp.route('/login', methods=['GET', 'POST'])
def login():
    SAML_ENABLED = current_app.config.get('SAML_ENABLED')

    current_app.logger.debug( current_app.config.get('SAML_ENABLED') )
```

rebuild the container:

`
docker build -t tag .
`
## 1. Test
create a `docker-compose.yml`, where I set the SAML_ENABLED to True.

```
version: "3"

services:
  app:
    image: tag:latest
    volumes:
      - "/root/cert:/app/cert"
    container_name: powerdns_admin2
    ports:
      - "80:80"
    logging:
      driver: json-file
      options:
        max-size: 50m
    environment:
      - SAML_ENABLED=True
```
(... and other environments that are not relevants at now)

`docker-compose up` and I loaded the `/login` page:

```
powerdns_admin2 | [2020-03-31 09:59:00,583] [index.py:147] DEBUG - True
```
That is OK, it is the expected operation.

## 2. Test

I switched the SAML_ENABLED to False in the compose file:
```
      - SAML_ENABLED=False
```

`docker-compose up` and I loaded the `/login` page:

```
powerdns_admin2 | [2020-03-31 10:02:02,676] [index.py:147] DEBUG - True
```

## 3. Test

I removed the SAML_ENABLED environment from the compose. In this case:

```
powerdns_admin2 | [2020-03-31 10:02:50,235] [index.py:147] DEBUG - False
```

## 4. Test

I removed from the `configs/docker_config.py` file the SAML_ENABLED value from the legal_envvars_bool array

`
docker build -t tag .
`

I inserted again this envrionment:
```
      - SAML_ENABLED=False
```
`docker-compose up` and I loaded the `/login` page:

```
powerdns_admin2 | [2020-03-31 11:47:13,738] [index.py:147] DEBUG - False
```

I experienced this with the SAML_ASSERTION_ENCRYPTED envrionment too.

I think this is a similar bug, that they wrote in this issue: 
https://github.com/docker/compose/issues/1788

ngoduykhanh: Thanks for the detailed debugging log, you are right. It has a bug in the docker boolean config parsing. I fixed it on commit b2d72d6603ea5915f62bc59590ac4af7c450e554. Can you try again with this PR? Now I think you can use `settings['security']['wantAssertionsEncrypted'] = current_app.config.get( 'SAML_ASSERTION_ENCRYPTED', True)`

Atisom: Thank you, I added the update! :)

ngoduykhanh: LGTM. Merging your PR. Thanks :)

PR #691, opened 2020-03-27
Add docker secrets support

athalonis:
Can you add this request, it enables support for docker secrets. This gives a more secure option of storing credentials and pass them to containers.


ngoduykhanh:
LGMT. It is a helpful feature. Mering your PR. Thanks.

PR #698, opened 2020-04-16
Support only having one auth method enabled 

Marcus-IXAU:
Currently you can't disable the local db auth if you are using an oauth provider (for example Microsoft Azure).

This adjust the validation to require at least one enabled authentication provider, and ends up being a little noise for a commit


ngoduykhanh:
LGTM. Merged your PR. Thanks for your contribution.

PR #700, opened 2020-04-17
Offline mode and searchable inputs

RoeiGanor:
Added some changes so powerdns-admin can work in an offline environment (without using google or gravatar.com).

In order to run on offline mode, set `OFFLINE_MODE` variable to `True`, and in order to use external services set `OFFLINE_MODE` to `False`:
```
 environment:
      - OFFLINE_MODE=True
```

In order to change favicon/gravatar, replace them in static/img/ like this:
docker-compose:
```
 volumes:
      - "<path-to-grav-icon>:/app/powerdnsadmin/static/img/gravatar.png"
      - "<path-to-favicon>:/app/powerdnsadmin/static/img/favicon.png"
```


RoeiGanor:
In order to find users easily when u have large environment with a lot of users, i have added search bar on top of the users list in the account edit menu, this will make the process of adding user to account much easier.
Screenshot from 2020-04-22 20-02-01 Screenshot from 2020-04-22 20-02-06

configs/development.py
```diff
@@ -6,6 +6,7 @@
 SECRET_KEY = 'e951e5a1f4b94151b360f47edf596dd2'
 BIND_ADDRESS = '0.0.0.0'
 PORT = 9191
+OFFLINE_MODE = True
```
ngoduykhanh, line 9: I think OFFLINE_MODE should have the False value by default. So we can keep the original state for the existing PDA users when they do upgrading.

ngoduykhanh: `OFFLINE_MODE` should be available in the `default_config.py` file as well.

configs/docker_config.py
```diff
@@ -44,6 +44,7 @@
     'SAML_LOGOUT',
     'SAML_LOGOUT_URL',
     'SAML_ASSERTION_ENCRYPTED',
+    'OFFLINE_MODE'
```
ngoduykhanh, line 47: `OFFLINE_MODE` has boolean value, it should be in `legal_envvars_bool` as well.

powerdnsadmin/__init__.py
```diff
@@ -100,4 +100,9 @@ def inject_setting():
         setting = Setting()
         return dict(SETTING=setting)
 
+    @app.context_processor
+    def inject_mode():
+        setting = app.config['OFFLINE_MODE'] if 'OFFLINE_MODE' in app.config else 'False'
```
ngoduykhanh, line 105: We can replace this line with
```
setting = app.config.get('OFFLINE_MODE', False)
```

We should treat it as a boolean value instead of a "string" one. Otherwise, the variable you set in the docker-compose environment won't work.

ngoduykhanh:
Thanks for the PR. I have a few in-line code comments which you can see above.
The search bar feature is nice! Do you think you can add it to the Domain Access Control page as well?
Also, please rebase your commits.

RoeiGanor:
@ngoduykhanh fixed the issues, added search inputs in domain edit, and rebased the code.

ngoduykhanh:
Awesome! I've merged your PR. Thanks for your contribution.

PR #703, opened 2020-04-20
Use sqlite by default in docker

kaechele:
This allows for easy setup with no extra configuration.
Also update docs to reflect the new and easy way to run PowerDNS-Admin.

Signed-off-by: Felix Kaechele <felix@kaechele.ca>


ngoduykhanh:
LGTM. Merging your PR. Thanks @kaechele

Issue #713, opened 2020-05-02
OpenID Connect name claim issue


maximumG:
First thanks for your work on this PowerDNS admin interface.

I setup OIDC login using the OpenID Connect OAuth tab with Keycloak (settings at the end) but ran into an issue regarding the name claim that is used create a new User in PDNS-Admin.

From the official OIDC specification, the `name` claim is the [full name of a user](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims). However PDNS-Admin is using this claim as if it was the [firstname only](https://github.com/ngoduykhanh/PowerDNS-Admin/blob/master/powerdnsadmin/routes/index.py#L289). As a result the `User` is created with a wrong Firstname (full name in it) and Lastname (empty).

I would rather use the `given_name` and `family_name` claims instead of the name. I can create a pull request with the new code if you'd like ?

OIDC settings with keycloak to reproduce

* Scope: email
* API URL: http://my.kc.home:8080/auth/realms/master/protocol/openid-connect/
* Token URL: http://my.kc.home:8080/auth/realms/master/protocol/openid-connect/token
* Authorize URL: http://my.kc.home:8080/auth/realms/master/protocol/openid-connect/auth

ngoduykhanh:
PR merged. Closing the issue.

PR #714, opened 2020-05-02
Fix OpenID Connect wrong claim name

maximumG:
Linked to issue #713

### DIFF
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -286,8 +286,8 @@ def login():
     if 'oidc_token' in session:
         me = json.loads(oidc.get('userinfo').text)
         oidc_username = me["preferred_username"]
-        oidc_givenname = me["name"]
-        oidc_familyname = ""
+        oidc_givenname = me["given_name"]
+        oidc_familyname = me["family_name"]
         oidc_email = me["email"]
 
         user = User.query.filter_by(username=oidc_username).first()


RoeiGanor:
tested it in my setup. When a user is created using oidc, the first name is the full name and there is no last name.
First I checked what the oidc userinfo return from the oidc and here is the given object:
`{'sub': 'xxx-111-yyy', 'email_verified': False, 'name': 'test3 test3', 'preferred_username': 'test3', 'given_name': 'test3', 'family_name': 'test3', 'email': 'test3@test3.com'}`
As we can see, the given name and family name are both populated with the needed value.
Then I deleted the user and recreated it using the proposed fix. The user created as expected.
I can also confirm that there were no impact on how the names (first and last name) are displayed in the UI.

terbolous:
Should these variable names maybe be configurable? I can imagine that different oidc providers may name them differently

maximumG:
> Should these variable names maybe be configurable? I can imagine that different oidc providers may name them differently

If they follow the OIDC specification, then it won't be different. Anyway, in case a provider is using non-standard claims, I guess it could be easy to add some form inputs with default to given_name and family_name.

ngoduykhanh:
I think we better have those claim values configurable for flexibility. In my experience, the claims name may differ among providers. Even in self-hosting OIDC, we can have custom claims.

maximumG:
Would it be possible to merge this pull request as well as #717 please ?

ngoduykhanh:
I've merged the #717 . Closing this PR. Thank you, everyone.

PR #742, opened 2020-05-22
Support authenticating using REMOTE_USER environment variable

nfantone:
- If new `REMOTE_USER_ENABLED` config setting is set to `True`, a username can be provided through a `REMOTE_USER` environment variable, which will be used to login into PowerDNS as a "trusted user".
- A new `LOCAL` user will be created as well and used in subsequent requests. This mimics the existing _Basic Authentication_ flow.
- Allows easy integration with existing SSO providers, such as [CAS](https://github.com/apereo/mod_auth_cas).

New available settings (values shown below are examples, not defaults):

```py
# Whether to enable remote user authentication or not
# Defaults to False
REMOTE_USER_ENABLED=True

# If set, users will be redirected to this location on logout
# Ignore or set to None to avoid redirecting altogether
# Warning: if REMOTE_USER environment variable is still set after logging out and not cleared by
# some external module, not defining a custom logout URL might trigger a loop
# that will just log the user back in right after logging out
REMOTE_USER_LOGOUT_URL=https://my.sso.com/cas/logout

# An optional list of remote authentication tied cookies to be removed upon logout
REMOTE_USER_COOKIES=['MOD_AUTH_CAS', 'MOD_AUTH_CAS_S']
```

This is my first PR here and I'm not really familiar with the overarching architecture of PDNS, so there might be a better/different approach to take here. I'll be happy to revisit the implementation if that's indeed the case.

### DIFF
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -377,8 +377,9 @@ def clear_session():
     session.pop('github_token', None)
     session.pop('google_token', None)
     session.pop('authentication_type', None)
-    session.clear()
+    session.pop('remote_user', None)
     logout_user()
+    session.clear()
 
 
 def signin_history(username, authenticator, success):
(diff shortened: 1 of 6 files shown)


powerdnsadmin/routes/index.py
```diff
@@ -434,7 +435,30 @@ def logout():
...
+            .format(current_user.username, remote_user_logout_url))
+        # Warning: if REMOTE_USER environment variable is still set and not cleared by
+        # some external module, not defining a custom logout URL will trigger a loop
+        # that will just log the user back in right after logging out
+        res = make_response(redirect(remote_user_logout_url.strip()))
+
+        # Remove any custom cookies the remote authentication mechanism may use
+        # (e.g.: MOD_AUTH_CAS and MOD_AUTH_CAS_S)
+        remote_cookies = current_app.config.get('REMOTE_USER_COOKIES')
+        for r_cookie_name in utils.ensure_list(remote_cookies):
```
ngoduykhanh, line 457: `ensure_list` function is missing.

ngoduykhanh:
Hi @nfantone ,

Thank you for submitting the PR.
To be honest I don't really understand the use case of this "remote user". If you can explain or contribute some information under `docs/` it would be great.

Btw, I added some comments and review for your PR :).

powerdnsadmin/routes/index.py
nfantone: Silly omission. Good catch. Added back.

nfantone:
@ngoduykhanh Config values added to requested sources.

nfantone:
> To be honest I don't really understand the use case of this "remote user".

Remote authentication is a widely used login mechanism that allows applications to delegate login responsibilities to an external provider. It's based on the principle that a third party product module handles the authentication of a user on the web server and that the information is passed on to the application in a server environment variable, typically named `REMOTE_USER`.

It has extensive support in pretty much any web framework out there, including [django](https://docs.djangoproject.com/en/3.0/howto/auth-remote-user/) and [Flask](https://tedboy.github.io/flask/generated/generated/flask.Request.remote_user.html).

From the [django docs](https://docs.djangoproject.com/en/3.0/howto/auth-remote-user/#authentication-using-remote-user):

> This type of authentication solution is typically seen on intranet sites, with single sign-on solutions such as IIS and Integrated Windows Authentication or Apache and `mod_authnz_ldap`, CAS, Cosign, WebAuth, `mod_auth_sspi`, etc.


Happy to add these bits above to `docs/` if you feel like that's necessary.

ngoduykhanh:
Hi @nfantone ,

Thank you for your clarification and fixes. I understand it now :-).
Your changes look good to me. Merging your PR...

PR #738, opened 2020-05-19
Add url_for() for static assets

ymage:

PR #763, opened 2020-06-24
Fix gravatar offline mode

RoeiGanor:
After #738 gravatar broke in offline because of incorrect templating at:
https://github.com/ngoduykhanh/PowerDNS-Admin/blob/4e63f8380bd6daa3fe5a651ee6d600be64f6dfd8/powerdnsadmin/templates/base.html#L32


ngoduykhanh:
Merged. Thanks for the fix.

PR #773, opened 2020-07-03
Add Account creation/permission handling based on Azure oAuth group membership

terbolous:
Messed up with some merges in the previous PR.

This will create a 1:1 mapping between Azure AD groups a user is member of and PowerDNS-Admin Accounts.
If the account exists, membership/permissions will be checked and if necessary permission granted.

A user can be member of multiple groups, and to avoid creating a lot of unnecessary groups there is support for regex matching (ie. you can use a prefix or similar)


ngoduykhanh:
Thank you for your contribution. I didn't have a chance to look at this earlier. Merging now.

PR #793, opened 2020-07-23
Added missing \ in docker run

vmag:
There is a missing \  in README docker run cmd.


ngoduykhanh:
Oh, I didn't see your PR earlier. Someone else fixed it. Thanks for your contribution btw.

PR #800, opened 2020-08-06
OIDC User and Account management during login

LordGaav:
This PR adds the following:

1. Update User details on every login.
2. If `oidc_oauth_account_name_property` and `oidc_oauth_account_description_property` are set, automatically create or update an Account during login, and add the User to it.
3. Allows setting of `oidc_oauth_logout_url` to automatically redirect the user to an IdP logout URL, for single logoff .

# Our use case

We store all our user information in an IdP, and can match a User to an Account using just the information returned from the userinfo response. For instance:

```json
{
  "given_name": "Mary",
  "family_name": "Lamb",
...
  "my-customer-id": "customer123",
  "my-customer-name": "Customer 123"
}
```

Setting `oidc_oauth_account_name_property` and `oidc_oauth_account_description_property` to 'my-customer-id' and 'my-customer-name' respectively will make sure that an Account is created (or updated) during login. This allows domains to automatically be mapped to this Account during the next domain sync.

### DIFF
--- a/powerdnsadmin/models/setting.py
+++ b/powerdnsadmin/models/setting.py
@@ -90,6 +90,7 @@ class Setting(db.Model):
         'oidc_oauth_api_url': '',
         'oidc_oauth_token_url': '',
         'oidc_oauth_authorize_url': '',
+        'oidc_oauth_logout_url': '',
         'oidc_oauth_username': 'preferred_username',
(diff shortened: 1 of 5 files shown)


ngoduykhanh:
LGTM. Merging.  Thank you for your contribution.
It sounds like we need the docs as OIDC config fields keep increasing now :)

PR #801, opened 2020-08-06
Implement account update method

LordGaav:
Allow syncing of all known accounts from PowerDNS, in the same way that Domain().update() does for domains.

# Our use case
We use OIDC as our login method (see #800 for more details). Using the new method in this PR we can pre-create all known Accounts for all domains, so that when the user logs in, he is automatically matched to the right Account and can manage his domains immediately. We can also now sync all Accounts and Domains in one job:

```python
import sys
import logging

from powerdnsadmin import create_app
from powerdnsadmin.models.account import Account
from powerdnsadmin.models.domain import Domain
from powerdnsadmin.models.setting import Setting

app = create_app()
app.logger.setLevel(logging.INFO)

with app.app_context():
    status = Setting().get('bg_domain_updates')

    ### Check if bg_domain_updates is set to true
    if not status:
        app.logger.error('Please turn on "bg_domain_updates" setting to run this job.')
        sys.exit(1)

    Account().update()
    Domain().update()
```

### DIFF
--- a/powerdnsadmin/models/account.py
+++ b/powerdnsadmin/models/account.py
@@ -98,7 +98,7 @@ class Account(db.Model):
         db.session.commit()
         return {'status': True, 'msg': 'Account updated successfully'}
 
-    def delete_account(self):
+    def delete_account(self, commit=True):
         """
         Delete an account
         """
@@ -107,7 +107,8 @@ class Account(db.Model):
 
         try:
             Account.query.filter(Account.name == self.name).delete()
-            db.session.commit()
+            if commit:
+                db.session.commit()
             return True
         except Exception as e:
             db.session.rollback()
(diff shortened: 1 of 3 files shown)


LordGaav:
Running into a puzzling error at the moment when calling `delete_account`:

```
[2020-08-06 14:54:38,498] [account.py:246] ERROR - Can not delete account from DB. DETAIL: 'Account' object has no attribute 'username'
```

ngoduykhanh:
The idea and code look good to me. Are they any change you want to add or I can merge it?

LordGaav:
> 
> 
> The idea and code look good to me. Are they any change you want to add or I can merge it?

It's been a while, let me check what the version of the code is we're running in production. Unfortunately I can only do that at the office, I'll get back to you next week.

LordGaav:
Just checked, we're still running this code in production with no issue, so you can just merge it as-is :+1:

ngoduykhanh:
Merging. Thank you for your contribution.

Issue #803, opened 2020-08-08
Link API key to account/user


RoeiGanor:
Currently, when creating a domain, you can link the domain to certain account and add users to ACL.
When you create an API key you grant access to domains.
Let's say i have a team working together and they have an account that has all of their domains, if I give them an API key and later add a domain to the account, I have to manually go and update they key so the team can work against the new domain.

I think it's a good idea to link the API  key to the domains that the account or user has, and this way, whenever account/user permissions are updated, their API key will be updated as well.

jbe-dw:
@RoeiGanor isn't it fixed with your last PR ? This issue could be closed.

PR #804, opened 2020-08-08
Add API Key to the UI

RoeiGanor:
I recently came across the feature of API Keys. The feature was pretty hidden, under-documented, and unintuitive.
I think it will be great to add this API key feature to the UI so the admin can create/delete/update API Keys easily from the portal and not through curl or postman.

Currently, this feature is implemented exactly like the API, where only admins can create keys.
As I mention in issue #803 I think we can extend this feature of API keys even more and make it more accessible and more easily used.

### DIFF
--- a/powerdnsadmin/routes/admin.py
+++ b/powerdnsadmin/routes/admin.py
@@ -1,6 +1,7 @@
 import json
 import datetime
 import traceback
+from base64 import b64encode
 from ast import literal_eval
 from flask import Blueprint, render_template, make_response, url_for, current_app, request, redirect, jsonify, abort, flash, session
 from flask_login import login_required, current_user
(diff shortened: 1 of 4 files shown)


ngoduykhanh:
LGTM. Thank you for your contribution. This is what I wanted to add before :-).
Merging...

HOSTED-POWER:
Hi,


How can we use this api ? We're very interested, do you have any sample commands / documentation? :)

RoeiGanor:
You can now create an api in the ui and associate it to domains. Then you can use the OpenAPI at /swagger to search for the api you want, and then add your key to X-API-Key header.
Like I said in the PR. this feature is really under documented, but you can use this as a reference:
https://github.com/ngoduykhanh/PowerDNS-Admin/blob/master/docs/API.md

HOSTED-POWER:
Ok thanks a lot @RoeiGanor !

One thing strange left, most api's take about 10-15 seconds to complete, this is rather long, even for simple api calls :/ :)

PR #814, opened 2020-09-09
do not fail with wildcard PTRs

TomHetmer:
Hi, 

fixes a bug where PowerDNS-Admin will fail with 500 when

- zone is ipv6
- pretty ptr is enabled
- zone contains wildcards


ngoduykhanh:
LGTM. Merged. Thank you for your contribution.

PR #817, opened 2020-09-16
Update README.md

virtualox:
Missing \ in docker run script at option 1.


ngoduykhanh:
Merging...Thanks for fixing it :)

PR #818, opened 2020-09-19
USER instruction enabled in Dockerfile

birkb:
- avoid running gunicorn as root user
- use setcap to allow gunicorn starting with a privileged port
- write permission only on /app/configs folder
- su command removed from entrypoint.sh, because it runs as pda user
- tested with `docker build -f docker/Dockerfile -t powerdns-admin:master .`


ngoduykhanh:
LGTM. Merged. Thanks.

PR #819, opened 2020-09-27
fix Dockerfile and typo in README.md

kaechele:
We need to ensure we own our /data dir on every launch since ownership
may change if a new volume is created and/or mounted

Also add a missing backslash in the readme

Signed-off-by: Felix Kaechele <felix@kaechele.ca>


ngoduykhanh:
Some previous PRs already updated the README and changed the Dockerfile also. Thank you for your PR anyway.

Issue #844, opened 2020-12-04
[API] some change proposals to be usable by terraform powerdns provider with user api key. (feature request)


WhatshallIbreaktoday:
Hello, 

Played a little bit with powerdns-admin and terraform powerdns provider ([https://github.com/terraform-providers/terraform-provider-powerdns]).

I was thinking on the following use-case (allowing other teams to automate without giving them Admin api key) : 
- distribute Api-keys to other teams (with limitied user rights)...
- Create the zones myself (with admin rights)...
- allow other teams to use terraform powerdns provider with their user api key to register their records themselves. (obviously zone creation will not work for them in this case).

Powerdns terraform Provider first requests : api/v1/servers/   and api/v1/servers/<server_id>  to retrieve Powerdns version and API versions...
So my proposal is to change API routes to let User API keys make this API calls, with of course letting only admins change zones...
But I have some doubts about this :  
Is it feasible without creating some security breach?  (I'm not so familiar with python applications).

One more thing : Terraform Powerdns Provider allows to  put trailing dot in zone name.. which is not allowed in the apikey_can_access_domain function  => adding .rstrip(".") when retrieving zone_id solves the issue and passes the requests correctly to powerdns. 

Great work anyway and have a nice day.

jbe-dw:
Hello,

I tested it weeks ago and came accross the same issue. You need an admin role to access the cited endpoints. This point is crucial. It would be so nice to fix it quickly because here at DW we deploy PowerDNS because of its API, but PDNS team doesn't want to manage RBAC so we need a frontend API to manage it.

I am deploying PowerDNS Admin and the improvments on API in the latest release is a good point. But with this permission issue, right now I have to install another bit of software to filter out API calls that is compatible with terraform.

I guess your project would be more famous if you can brand that it provides a PowerDNS Api frontend with RBAC control compatible with terraform PowerDNS Provider.

Thank you, i'm interested in it. I may do a PR but I'm quite busy to dive into your code now.

PR #845, opened 2020-12-04
Tweaks to allow  user apikey usage with powerdns terraform provider

WhatshallIbreaktoday:
As described in  #844, please find the tweaks I made to allow User api keys to use Terraform Powerdns Provider to create records. (zone creation will obviously not work with user api key.  You have to provide admin Key for that).

Please review this proposal (not sure about creating security issues by opening /servers and /servers/<server_id> (GET only) to non admin users .. (but still registered api key)...

Best Regards,

### DIFF
--- a/powerdnsadmin/decorators.py
+++ b/powerdnsadmin/decorators.py
@@ -236,7 +236,7 @@ def apikey_can_access_domain(f):
         apikey = g.apikey
         if g.apikey.role.name not in ['Administrator', 'Operator']:
             domains = apikey.domains
-            zone_id = kwargs.get('zone_id')
+            zone_id = kwargs.get('zone_id').rstrip(".")
             domain_names = [item.name for item in domains]
 
             if zone_id not in domain_names:
--- a/powerdnsadmin/routes/api.py
+++ b/powerdnsadmin/routes/api.py
@@ -913,7 +913,7 @@ def api_zone_forward(server_id, zone_id):
                                   created_by=g.apikey.description)
                 history.add()
         elif request.method == 'DELETE':
-            history = History(msg='Deleted zone {0}'.format(domain.name),
+            history = History(msg='Deleted zone {0}'.format(zone_id),
                               detail='',
                               created_by=g.apikey.description)
             history.add()
--- a/powerdnsadmin/routes/api.py
+++ b/powerdnsadmin/routes/api.py
@@ -919,15 +919,6 @@ def api_zone_forward(server_id, zone_id):
             history.add()
     return resp.content, resp.status_code, resp.headers.items()
 
-
-@api_bp.route('/servers', methods=['GET'])
-@apikey_auth
-@apikey_is_admin
-def api_server_forward():
-    resp = helper.forward_request()
-    return resp.content, resp.status_code, resp.headers.items()
-
-
 @api_bp.route('/servers/<path:subpath>', methods=['GET', 'PUT'])
 @apikey_auth
 @apikey_is_admin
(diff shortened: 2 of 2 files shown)


jbe-dw:
@ngoduykhanh It would be nice to review it quickly. Just 12 lines of change. Tested successfully.

ngoduykhanh:
As most of the use case of PDA is internal. I think allowing GET on the server config is OK for this feature for now.
The code looks good to me. @WhatshallIbreaktoday Do you have more changes want to add? Otherwise, I will merge it.

WhatshallIbreaktoday:
Nothing to add. I'm happy with this :)

ngoduykhanh:
Merging. Thanks for your contribution.

florianbeer:
What kind of information will this endpoint present back to the users?
If it is security relevant, I may need to block this on the webserver level in our installation.

ngoduykhanh:
@florianbeer , it exposes information about the PDNS config under `/servers` endpoints (Read-only). If it has sensitive information you want to hide, you can block that URL at your web server / API gateway level config.

florianbeer:
Yes, that's exactly my question: what is the information exposed at this endpoint? Can you please provide an example?

> As most of the use case of PDA is internal.

This is a pretty bold assumption! We are using the PowerDNS-Admin API to give customers the possibility to administer their zones, so I have to make absolutely sure there is nothing available they should not see.

If you have to make such drastic changes, at least issue a security advisory so admins don't deploy this version without being aware that suddenly additional information is visible.

ngoduykhanh:
@florianbeer In the new release of PDA, there will be notes for changes like this. The Administrator has to aware before upgrading.
Btw, since this change causes conflicts, we will try to make it works in different use cases, as much as possible. The possible solutions I can think of:
- Create another dedicated API for using with Terraform.
- Have options to allow/deny access from user-key.

Please raise another issue for this. Thanks.

WhatshallIbreaktoday:
Info exposed are described here : 
[https://doc.powerdns.com/authoritative/http-api/server.html]

On my test server, here is the output  (on a GET request), with a User API Key (not Administrator one).

curl --location --request GET 'https://redacted_myserver/api/v1/servers/localhost' \
--header 'X-API-KEY: redacted_myapikey' \

{
    "config_url": "/api/v1/servers/localhost/config{/config_setting}",
    "daemon_type": "authoritative",
    "id": "localhost",
    "type": "Server",
    "url": "/api/v1/servers/localhost",
    "version": "4.2.3",
    "zones_url": "/api/v1/servers/localhost/zones{/zone}"
}

when I want to get a specific config:

curl --location --request GET 'https://redacted_myserver/api/v1/servers/localhost/config' \
--header 'X-API-KEY: redacted_myapikey' \

I get : {"msg": "Not enough privileges"}

Hope it is clearer
Best regards,

jbe-dw:
@florianbeer 

What changes does the code do:

- Before the patch (0.2.3): /servers and any subfolders needed an admin api key, on each method, excepted the /servers/<server_id>/zones
- With the patch: any user is able to browse /servers, /servers/<server_id>, but /servers/<server_id>/config (which exposes the PowerDNS config) still needs an admin api key

What informations do we have on this two endpoints:
/servers and /servers/<server_id> expose the same thing
[{"config_url": "/api/v1/servers/localhost/config{/config_setting}", "daemon_type": "authoritative", "id": "localhost", "type": "Server", "url": "/api/v1/servers/localhost", "version": "4.3.1", "zones_url": "/api/v1/servers/localhost/zones{/zone}"}]

The PDNS version is exposed

@ngoduykhanh I don't think it's worth doing a separate api for terraform.

florianbeer:
> I don't think it's worth doing a separate api for terraform.

I concur, the only slightly "sensitive" information is the version string, which is ok from my point of view for customers who've already received an api key from us.

jbe-dw:
This setting can be shadowed directly in pdns itself by setting version-string to a custom text

PR #855, opened 2020-12-30
allow-server-side-sessions

jodygilbert:


jbe-dw:
@jodygilbert Hello

Does this patch solve the annoying disconnection issue (if my session times out i'm redirected to the login screen, then the first attempt gives a 403 error and I have to relog or just hit refresh button on the login screen) ?

jodygilbert:
This fixes the issue where the cookie created by Azure AD is larger than 4K, which causes pda auth to fail (you end up in an endless redirect loop). I will see if I can replicate the issue you experience and if this patch fixes it.

powerdnsadmin/__init__.py
```diff
@@ -78,6 +79,11 @@ def create_app(config=None):
         from flask_sslify import SSLify
         _sslify = SSLify(app)  # lgtm [py/unused-local-variable]
 
+    # Load Flask-Session
+    if app.config.get("SESSION_TYPE") is not None and app.config.get("SESSION_TYPE") != 'null':
```
ngoduykhanh, line 83: Can we just use
```python
app.config.get("SESSION_TYPE")
```
instead of
```python
app.config.get("SESSION_TYPE") is not None
```
And, why it cannot be `null` string? What is the possible input for this `SESSION_TYPE` env variable? Can we validate them?

configs/docker_config.py
```diff
@@ -47,7 +47,13 @@
...
     'OFFLINE_MODE',
     'REMOTE_USER_LOGOUT_URL',
-    'REMOTE_USER_COOKIES'
+    'REMOTE_USER_COOKIES',
+    'SESSION_TYPE',
+    'SESSION_REDIS',
+    'SESSION_MEMCACHED',
+    'SESSION_MONGODB',
+    'SESSION_SQLALCHEMY',
+    'SESSION_SQLALCHEMY_TABLE'
```
ngoduykhanh, line 56: Each session type here requires an extra driver for making the connection. Can you add a [docs](https://github.com/ngoduykhanh/PowerDNS-Admin/tree/master/docs) for it? An addition `requirements-extras.txt` also good I think.

ngoduykhanh (changes requested):

powerdnsadmin/__init__.py
jodygilbert: Yep, your suggestion for app.config.get("SESSION_TYPE") makes perfect sense.

The only reason I excluded null was that is appears the flask-session requires a SECRET_KEY if you set the session_type to null (or at least my container kept getting errors about the session_key when the SESSION_TYPE was set to null.  
I had considered including a new variable for SECRET_KEY or using the existing SECRET_KEY variable, but I didn't want to break existing builds by doing this.  Happy to do either though really.

configs/docker_config.py
jodygilbert: that's a good point about drivers, I hadn't considered that.  I was mainly thinking that if filesystem was added, it made sense to add the others.  However, I'm probably not the best person to test the other options, I will remove them.  If they are required they can be added later.

jodygilbert:
> @jodygilbert Hello
> 
> Does this patch solve the annoying disconnection issue (if my session times out i'm redirected to the login screen, then the first attempt gives a 403 error and I have to relog or just hit refresh button on the login screen) ?

I'm afraid this doesn't fix the issue with the 403 error, I'm not sure why that happens.

jbe-dw:
I guess something mismatch in the user session. I'll take a look at this because it bores me a lot, even though i'm a lot on the API now.

powerdnsadmin/__init__.py
jodygilbert: I've removed the untested session type and used a FILESYSTEM_SESSIONS_ENABLE configuration instead, there is validation for this being boolean.

ngoduykhanh:
Merging, thanks @jodygilbert

Issue #848, opened 2020-12-16
Feature Request: Add a configuration API endpoint with initial admin and/or apikey setup


itbane:
Hi,

by trying to automate the deployment of PowerDNS-Admin, I noticed some showstoppers in regards of API usage. I have a couple of suggegstions that would allow for a much easier setup.

* Allow the API-key, created as in [1], to be used on all endpoints (either exclusively or parallel with the basic auth)
* Allow user registration via a new API endpoint to allow a hands-free initial installation
* Allow PDNS-Admin configuration via API (PDNS settings, Authentication, etc.)

As I am not a python developer, I think I can't contribute any actual code for this. If I can support in another way to help further this topic, please let me know.

KR, itbane

[1] https://github.com/ngoduykhanh/PowerDNS-Admin/blob/master/docs/API.md

jbe-dw:
Hello,

We have a plan to implement such things later. A configuration endpoint, and at least another one to create the initial admin.
I'm not quite sure if we want to allow the API Key for all endpoints (/powerdnsadmin) but we will decide it then.

Regards

jbe-dw:
Hello,

This request is stale for too long. Regarding our policy to cleanup this project, we are closing it. As this is a Feature Request, it will be listed in #1178

Regards
PowerDNS Admin Team
See #1153 for the policy

Issue #847, opened 2020-12-10
Api key authentication consistence


alxhlz:
Hey guys,

I was trying to fetch all zones of a user via api and found out, that this is only possible via `/pdnsadmin/zones` endpoint. Sadly this endpoint has to be authenticated by basic auth and not via api key.

Now my question is, why no consistence? Is there a particular reason for that?

Regards,
Alex

jbe-dw:
Hello,

With the latest version (master, maybe 0.2.4), users can list their zones via the /server/../zones endpoint.

Regards

PR #861, opened 2021-01-07
feat: Allow sync domain with basic auth

jbe-dw:
#847 #848 
Hello @ngoduykhanh 

I have written 2 ansible modules to drive your API: one for managing accounts, one for managing users. I want to show on the output the domains that are associated to the account. 

I create domains through Terraform, so I need to browse the sync_domain endpoint to refresh existing domains so the account is populated with associated domains.

The problem is that your API is split in two parts: 

* The management endpoints in /pdnsadmin is accessed with basic auth
* The PowerDNS proxied endpoint is accessed with an API Key
* The sync_domain endpoint is surprisingly also protected by the API Key

To keep my module consistent (the ones to manage PDNS Admin) I use a traditional user/password setup to browse PDNS Admin API with basic auth. We are also supposed to do so to create the Administrator API key after install. 

I won't ask a user to first create a key, to then update their ansible playbook to add the key to be able to browse /sync_domain endpoint in order to retrieve fresh informations.

This patch implements a generic decorator that allow a fallback to basic on this specific endpoint. As no permission is tested it should be fine. It's an easy option to answer some issues that ask to add api key support on the admin enpoints as well, though some work is intended to replace basic authentication routes permission checking.

What do you think ?

### DIFF
--- a/powerdnsadmin/decorators.py
+++ b/powerdnsadmin/decorators.py
@@ -291,3 +291,13 @@ def dyndns_login_required(f):
         return f(*args, **kwargs)
 
     return decorated_function
+
+def apikey_or_basic_auth(f):
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        api_auth_header = request.headers.get('X-API-KEY')
+        if api_auth_header:
+            return apikey_auth(f)(*args, **kwargs)
+        else:
+            return api_basic_auth(f)(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


ngoduykhanh:
I think, ideally, the API authentication should be done via an API key. There PDA API needs to be improved also. However, I will merge this PR for now to solve your issue. Thanks.

Issue #856, opened 2021-01-04
API permission problem


Caez83:
Hi,

An API user can list all zones even if only one domain is listed on the ACL
On the web interface the same user can see only his domain. 

A simple test :
curl -sSfL --stderr - --request GET --header 'Content-Type: application/json' --header 'X-API-Key: API_KEY' --data '' https://dnsadmin/api/v1/servers/localhost/zones

RoeiGanor:
If you change the server from localhost to pdnsadmin then it should give you the correct domains, like it is written here:
https://github.com/ngoduykhanh/PowerDNS-Admin/blob/2917c47fd1c78194d3462687b3a31c17c42e308e/powerdnsadmin/routes/api.py#L960

Caez83:
Thank you but i think it's a security issue because any user can use localhost and list any domains 
So I've added 
if server_id == 'pdnsadmin' or server_id == 'localhost':

RoeiGanor:
Well, actually I don't think this is a security breach. Any unauthenticated user can dig/nslookup to a dns server and see what domains/records are configured there, let alone this is an authenticated api.
I think this is a misleading api, not a security problem.

jbe-dw:
I guess that you're expecting this behaviour because you're distributing API keys to limit managers view over the API. It's reasonable to think that you may not want the see all domains on this endpoint because it's a bit disturbing and inconsistent with the idea that a user could list any domain and get an error trying to access it.

In my setup I have set PDNS as a frontend to use API keys to limit user actions to a list of allowed domain. The changes will be made through Terraform to allow them to create A and PTR records when they pop up virtual instances on the delegated domains, so it uses the server/localhost endpoint.

In my opinion it should be consistent accross the two endpoints. If there's no regression I think It's the way to go if it only requires a tiny change in the code.

Caez83:
Yes, it's not a real security issue, but i would like the same view on web interface and on the API. 
It's more difficult to list and dig or nslookup each domain or they need to use an external service.

Some account are used to create acme records for wildcard letsencrypt domain and some script check all the domain right and give error if the domain is not allowed.
Generaly i'm creating an api access for each server which need to generate these entries. 
Before the last release, API and Web access gave the same view.

Generaly on a API view we can see only resources than we're authorized to see or to modify. 
So it's logic to permit Adminstrator or Operator to see any domain but an user don't need to see it. 

```
     if server_id == 'pdnsadmin':
[...]

PR #862, opened 2021-01-08
feat: limit zone list for users on servers endpoint

jbe-dw:
This patch enforces zone filtering on /servers/localhost/zones endpoint for API Keys with User role, thus limiting listing to allowed zones. It returns PDNS content untouched.

Response to #856


ngoduykhanh:
LGTM. Merging. Thank you!

PR #863, opened 2021-01-11
fix: user jsonify to set response headers to json

jbe-dw:
Hello @ngoduykhanh 

I tried to make a go client from your swagger spec with go-swagger tool. From the description, all returned data from the API is application/json, but headers were not set and defaulted to text/html.

The client raise an error because it expects json. I updated all json.dumps() to use Flask's jsonify() which correctly sets headers.

If there is any particular reason let me know.

Regards


ngoduykhanh:
Yes, using `jsonify` is better. Merging your PR. Thank you!

PR #864, opened 2021-01-14
feat: Implement apikeys/<id> endpoint from swagger spec.

jbe-dw:
@ngoduykhanh 

Hello,

The swagger specs defines a pdnsadmin/apikeys/<key_id> endpoint that was not implemented (it falls back on the /apikeys which return a list of keys instead of a specific key).

Added here


ngoduykhanh:
Thanks for your contribution. Merging the PR.

PR #877, opened 2021-01-29
Improve account creation/permission handling based on Azure oAuth group membership 

jodygilbert:
This reduces the number of graph api's calls by only using one call for all the groups, instead of a call for each group, which massively improves login time.  It also fixes the iteration through the list of groups.


jodygilbert:
I've just noticed this isn't pulling all groups if the user is a member of too many, I'm working on paging through the results now

jodygilbert:
paging now works, I tested by going down to 20 results per page

ngoduykhanh (approved):

ngoduykhanh:
Merging. Thank you for your contribution.

PR #878, opened 2021-02-01
upd: improve user api

jbe-dw:
@ngoduykhanh 

Hello

This PR include multiple changes:
- it changes the output of the user endpoint, so it's in line with the last changes to apikeys and accounts
- it changes the user/account link endpoints to a more logical path accounts/<account_id>/users/<user_id> (the old pattern is not removed but swagger definition has been updated to match the new one)
- it adds account information to users (if you get one, not the list)
- it fixes a bug I introduced in #874
- it introduce a cleaner way to output the results (using lima library to not output json object as a collection instead of tricking jsonify by giving the generated list first item as argument)

~~I need some help because I'm not comfortable with SQLAlchemy. I'd like to add an account property to the User class. I have tried to initialize the accounts field in the init() of the class with the get_accounts() function but it's not working. I do it in the api later but it feels ugly, I'd prefer to have it initialized itself whenever a user is instanciated.~~

This changes makes my terraform provider functional. A bit of work is needed though to correctly handle error codes and output (i.e. trying to create an object with an existing name should not trigger 500 Internal Server Error but 403 Forbidden or 409 Conflict )

Thank you for your time !

### DIFF
--- a/powerdnsadmin/lib/errors.py
+++ b/powerdnsadmin/lib/errors.py
@@ -93,6 +93,15 @@ class AccountCreateFail(StructuredException):
         self.name = name
 
 
+class AccountCreateDuplicate(StructuredException):
+    status_code = 409
+
+    def __init__(self, name=None, message="Creation of account failed"):
+        StructuredException.__init__(self)
+        self.message = message
+        self.name = name
+
+
 class AccountUpdateFail(StructuredException):
     status_code = 500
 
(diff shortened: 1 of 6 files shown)


jbe-dw:
It should be fine now

jbe-dw:
@ngoduykhanh  ?

ngoduykhanh:
LGTM. Thanks for your PR @jbe-dw . And sorry for the delay. I was super busy these days 😅

jbe-dw:
@ngoduykhanh I've seen that you just change your job, I can get it !

PR #597, opened 2019-12-08
Support IDN domain

ngoduykhanh:
As discussed in #556 
This PR contains basic support for IDN. For now, it is just for allowing adding the IDN domain and displaying them in the pretty format.

Some more adjustment can be done:
- Allow IDN domain searching
- Display the domain name in both formats to the user

Screen Shot 2019-12-08 at 14 33 48

PR #879, opened 2021-02-01
Add punycode (IDN) support

r-daneel:
Add support for punycode (IDN) domains and records.

This is built upon the original [PR#597](https://github.com/ngoduykhanh/PowerDNS-Admin/pull/597).

- domains and records now always show up pretty
- CNAME records can be pointed at IDN names
- IDN shows up correctly even when only the TLD is an IDN

- Search is still not working (will require smart adjustments that probably go deeper into how this works for now)
- Display of both idn encoded and non-idn-encoded values does not seem to be of much use for now (maybe as a tool-tip ?)
- JS modals that display domain/record names still do not show the pretty version yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ A PowerDNS web interface with advanced features.
 - DynDNS 2 protocol support
 - Edit IPv6 PTRs using IPv6 addresses directly (no more editing of literal addresses!)
 - Limited API for manipulating zones and records
+- Full IDN/Punycode support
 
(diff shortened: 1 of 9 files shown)


mrsrvman:
is anybody home?

r-daneel:
If someone could review and merge it would be nice :-)

ngoduykhanh:
Merged it. Thanks for your contribution :)

Issue #865, opened 2021-01-16
Letsencrypt and lego error - /api nout found


Cyr-ius:
The registration of certificates via the LEGO module is impossible because the path http: //xxxxx/api does not return the API version as can PowerDNS on its administration port

Is it possible to modify the reverse proxy to return this information ?

Syntax3rror404:
Use a nginx or traefik reverse proxy ...

jbe-dw:
#1206 (replaces #1115) has been merged to master.

Please use master branch or wait for next release.

Feel free to reopen if you still have issue with this.

PR #902, opened 2021-03-10
Change api url_prefix to add LEGO client compatibility

WhatshallIbreaktoday:
Hi.
This PR proposes a way for LEGO client to be able to get access to powerdns /api endpoint by modifying Powerdns-Admin api code like this :

- Move url_prefix to /  (instead of /api/v1/)
- Add /api/v1/ to all existing api routes.
- Add /api  (without ending / ) to allow Lego client to get powerdns api version.

Tested and working with patched LEGO client (a PR will also be proposed there) to add json content-type when performing PATCH requests.  (missing for now, not an issue when request addresses Powerdns directly, but blocked request by powerdns-admin).

Solves #865

Did not change swagger file, since the PR allows only to proxy request to Powerdns.. and leaves Powerdns-Admin Api unchanged.

If you have a better way to perform this. Please let me know.
B R,

### DIFF
--- a/powerdnsadmin/routes/api.py
+++ b/powerdnsadmin/routes/api.py
@@ -30,7 +30,7 @@ from ..decorators import (
 import random
 import string
 
-api_bp = Blueprint('api', __name__, url_prefix='/api/v1')
+api_bp = Blueprint('api', __name__, url_prefix='/')
 
 apikey_schema = ApiKeySchema(many=True)
 domain_schema = DomainSchema(many=True)
@@ -169,7 +169,7 @@ def before_request():
             }))
 
 
-@api_bp.route('/pdnsadmin/zones', methods=['POST'])
+@api_bp.route('/api/v1/pdnsadmin/zones', methods=['POST'])
 @api_basic_auth
 @api_can_create_domain
 def api_login_create_zone():
@@ -222,7 +222,7 @@ def api_login_create_zone():
     return resp.content, resp.status_code, resp.headers.items()
 
 
-@api_bp.route('/pdnsadmin/zones', methods=['GET'])
+@api_bp.route('/api/v1/pdnsadmin/zones', methods=['GET'])
 @api_basic_auth
 def api_login_list_zones():
     if current_user.role.name not in ['Administrator', 'Operator']:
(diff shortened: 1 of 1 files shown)


AgentTNT:
Where can we find the patched LEGO client that goes with this? Currently using v2.4.0 of the ACME provider with Terraform.

I've been trying to get terraform to work with both User and Administrator API keys via PowerDNS-Admin, and this PR was the closest thing that made it work. However it fails with return code 400 for the PATCH, which you do call out above. 

Even prior to this patch when using the PowerDNS-Admin as the PDNS API URL, it would continually error out since it couldn't find the "/api" version endpoint, and if specified manually in the URL it would double it up for return items like "api/v1/api/v1/servers/localhost..."

WhatshallIbreaktoday:
Hi, the Lego client that works for me is the following one : [https://github.com/go-acme/lego](https://github.com/go-acme/lego).

And the patch I was refering, is this one : 

`if method != http.MethodGet && method != http.MethodDelete {
		req.Header.Set("Content-Type", "application/json")
	}`

which is now included in a wider patch : [https://github.com/go-acme/lego/commit/83c626d9a1889fa499bc9c97bc2fdea965307002](https://github.com/go-acme/lego/commit/83c626d9a1889fa499bc9c97bc2fdea965307002)

Hope it helps,
Best regards

AgentTNT:
Thanks! I think my confusion stems from this PR you also submitted - https://github.com/ngoduykhanh/PowerDNS-Admin/pull/845/files

I think I was under the impression that the PDNS provider for acme was already working for the DNS challenge, but now I see it was only for creating new records via https://registry.terraform.io/providers/pan-net/powerdns/latest/docs. 

Do you have a working example of how to use the LEGO client directly in terraform for cert creation?

AgentTNT:
Scratch that. I see now this is a chicken and egg problem. The PR you mention was merged after the most recent release of LEGO, and therefore the current ACME Terraform module doesn't know about it. Once LEGO updates, then ACME Terraform updates, then everything should be happy!

jbe-dw:
Hello @WhatshallIbreaktoday 

There was some long period with no activity on the project in the last months. We are a new team of maintainers now that starts working on PDA. You opened it one year ago and I'm giving an answer just now.

Isn't #1115 adding the endpoint that LEGO miss and achieve the same goal in an easier fashion ? This PR implement a /api endpoint that informs the client about available API versions.

Can you get a look at it and try ? Lego client should have been updated since, hence we would like to know if the latest release are compatible with the PR I mentioned.

Regards

WhatshallIbreaktoday:
Hello, the PR you mentionned is far better than mine, and has been proposed after. So i think i will close this one as it answers perfectly.. in a better way. Thanks for your work guys ! You're awsome

Issue #915, opened 2021-03-30
Cannot delete domain.


falkowich:
Updated yesterday to:
https://hub.docker.com/layers/ngoduykhanh/powerdns-admin/latest/images/sha256-2d48412f49c9b1b76ed9430736a787ca6267ec8c63c3504c0bcc71b25a784a7a

When we try to delete domains this happens.

```
powerdns_admin | [2021-03-30 14:12:31,302] [domain.py:142] DEBUG - Traceback (most recent call last):
powerdns_admin |   File "/usr/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context
powerdns_admin |     self.dialect.do_execute(
powerdns_admin |   File "/usr/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 593, in do_execute
powerdns_admin |     cursor.execute(statement, parameters)
powerdns_admin |   File "/usr/lib/python3.8/site-packages/MySQLdb/cursors.py", line 206, in execute
powerdns_admin |     res = self._query(query)
powerdns_admin |   File "/usr/lib/python3.8/site-packages/MySQLdb/cursors.py", line 319, in _query
powerdns_admin |     db.query(q)
powerdns_admin |   File "/usr/lib/python3.8/site-packages/MySQLdb/connections.py", line 259, in query
powerdns_admin |     _mysql.connection.query(self, query)
powerdns_admin | MySQLdb._exceptions.IntegrityError: (1451, 'Cannot delete or update a parent row: a foreign key constraint fails (`pda`.`history`, CONSTRAINT `fk_domain_id` FOREIGN KEY (`domain_id`) REFERENCES `domain` (`id`))')
powerdns_admin | 
powerdns_admin | The above exception was the direct cause of the following exception:
powerdns_admin | 
powerdns_admin | Traceback (most recent call last):
powerdns_admin |   File "/app/powerdnsadmin/models/domain.py", line 138, in update
powerdns_admin |     self.delete_domain_from_pdnsadmin(domain_name, do_commit=False)
powerdns_admin |   File "/app/powerdnsadmin/models/domain.py", line 523, in delete_domain_from_pdnsadmin
powerdns_admin |     Domain.query.filter(Domain.name == domain_name).delete()
powerdns_admin |   File "/usr/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3899, in delete
powerdns_admin |     delete_op.exec_()
[...]

PR #916, opened 2021-04-02
Bugfix: delete history records when a domain is deleted

jodygilbert:
This fixes a bug I introduced, the history records for a domain are now deleted before the domain is deleted.

This will resolve issue #915


lvasiliadis (approved):
The fix worked as expected.
Thanks !

mzealey:
Would it be possible to do a (docker) release based on this as it's quite an annoying break at present?

jodygilbert:
My understanding is it just needs to be merged into the main project for the docker build to be updated.

herukrayz (approved):

Issue #919, opened 2021-04-13
Comments are not supported on all pdns backends


jyoung15:
It appears the fix in #676 is no longer working to remove blank comments.   With the LMDB backend, the following error appears when adding/editing records even if all comments are blank:

`Hosting backend does not support editing comments`

It appears the error is due to a change in the comment structure:
```json
            "comments": [
                {
                    "account": "",
                    "content": ""
                }
            ],

```

The previous fix checked if the comments value was empty, but now even with no comments the list is not empty.

I have a proposed fix and will submit a pull request.

PR #920, opened 2021-04-13
Delete blank comments.  Fix for ngoduykhanh/PowerDNS-Admin#919

jyoung15:
Proposed fix for backends that do not support comments, such as LMDB.  See ngoduykhanh/PowerDNS-Admin#919.


ngoduykhanh:
Thanks for providing the fix. Merging it.

PR #929, opened 2021-04-29
SAML improvements for Docker

icb-:
* Fixes a typo when managing user accounts through SAML assertions
* Adds a few more config. options from Docker env. (`SIGNUP_ENABLED`, `LOCAL_DB_ENABLED`, `LDAP_ENABLED`)
* Updates names for SAML key and cert files to match convention, so Docker secrets can be used


ngoduykhanh:
LGTM. Thanks for your contribution @icb-

PR #931, opened 2021-05-05
Emphasize importance of using a custom SECRET_KEY

jaylinski:
This project provides a default SECRET_KEY for signing session-cookies:
https://flask.palletsprojects.com/en/1.1.x/config/#SECRET_KEY

By using the default SECRET_KEY, everyone will be able to create valid session-cookies.
So users should be aware that it is very important to set a custom SECRET_KEY.

It would be even better if PowerDNS-Admin would refuse to start if the default SECRET_KEY is used.


ngoduykhanh:
Thanks for reporting and improve the docs @jaylinski .

Issue #939, opened 2021-06-02
Feature: Add background jobs to docker


steschuser:
PowerDNS-Admin supports the fetching of Domain and Account data in the background.
This results in a fast and snappy website.

For this to work, the setting "bg_domain_updates" has to be turned on in Settings-> Basic and the actual fetching of the data has to happen elsewhere.

One way is to request the route /domains-updater via cronjob or similar.
This needs a login and so exposes the password/token. It can also become a hassle if you have multiple instances running.

Alternative to that are the provided scripts update_zone and update_accounts. Those work fine and dont need additional settings as they can work with the PowerDNS-Admin database directly. This makes using the scripts superior to calling /domains-updater, imho.

However, this is impossible to setup with the provided Docker Container.
Therefor another Dockerfile is needed that enables the User to switch to Background jobs easily, without setting up external cronjobs.

TLDR:
Make "bg_domain_updates" work in Docker

steschuser:
#940

PR #940, opened 2021-06-02
Add background jobs to docker

steschuser:
#939 

This changes the files "update_zones.py" and "update_accounts.py" slightly, changing the error message to lower priority. It also moves them to the powerdnsadmin directory inside the docker image.

A new Dockerfile "Dockerfile.BackgroundJob" is created. This docker image differs from the stock one provided by including the s6-overlay https://github.com/just-containers/s6-overlay 
That way, a cron instance can be started next to PowerDNS-Admin.

Using that, the Docker Image will automatically update Zones and Accounts every 5 minutes if the "bg_domain_updates" setting is enabled.

TLDR:
Have Dockerfile with background jobs


steschuser:
#940

Issue #951, opened 2021-06-09
Feature: Allow users to remove domains


steschuser:
We want Users to be able to remove Domains, if the account belongs to them

PR #952, opened 2021-06-09
allow users to remove domain

steschuser:
#951

Adds a new route, new decorator, new setting and a new template to allow users to delete their domains

### DIFF
--- a/powerdnsadmin/routes/domain.py
+++ b/powerdnsadmin/routes/domain.py
@@ -10,7 +10,7 @@ from flask_login import login_required, current_user, login_manager
 
 from ..lib.utils import pretty_domain_name
 from ..lib.utils import pretty_json
-from ..decorators import can_create_domain, operator_role_required, can_access_domain, can_configure_dnssec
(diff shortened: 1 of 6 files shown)


powerdnsadmin/routes/domain.py
```diff
@@ -131,6 +134,47 @@ def domain(domain_name):
                            ttl_options=ttl_options)
 
 
+@domain_bp.route('/remove', methods=['GET', 'POST'])
+@login_required
+@can_remove_domain
+def remove():
+    if request.method == 'POST':
+        current_app.logger.debug(request.form)
+        domain_name = request.form['domainid']
```
RoeiGanor, line 143: The page is rendered only with domains you have access to, but during the deletion process there is no check to see if the domainid sent is a domain you own, which means, that any user can delete any domain by changing the form body

steschuser: Hello,

thanks.
Added a Check with new commit

Issue #670, opened 2020-03-02
Zone name as placeholder in domain template


prupert:
How can we use the zone name as a placeholder in domain template records in PowerDNS-Admin?

Previously this was possible by using `@` as contents in PowerDNS-Admin domain template records, but this now results in PowerDNS resolving records to a literal `\@.`

splog123:
I think that might be a PowerDNS thing. 

When I do INSERTs directly into a pdns database with "@" in the content of a record where it could belong, PowerDNS queries always return a literal \@

Seems like PowerDNS only accepts @ for $ORIGIN in the name field of a record ?

prupert:
So, how do we use domain names in a PowerDNS-Admin template? 

It seems the same question was also asked as part of a wider inquiry at https://github.com/ngoduykhanh/PowerDNS-Admin/issues/641

prupert:
Yes, we need to have the domain name available as (part of the) content, such as in CNAME records.

splog123:
Based on what I see, I don't think that works directly. We need ngoduykhanh to see this and tell us how to make it work for you.

prupert:
> Based on what I see, I don't think that works directly.

I believe it was previously possible before refactoring in https://github.com/ngoduykhanh/PowerDNS-Admin/commit/8ea00b948410bd6a63c971d068c7b64131ac8eda, but I am not entirely sure.

That said, I think there is a good use case for having a placeholder in the content of a domain template record. Other PowerDNS admin interfaces that I know of also have this feature, i.e. the placeholder `[ZONE]` in PowerAdmin.

prupert:
Perhaps @ngoduykhanh, if and when you have some time available, is this something that is still supported or will this require adding a new feature?

RobertIchim:
on file powerdns-admin/powerdnsadmin/models/record.py line 178 add on new blank line and 
record["record_data"] = record['record_data'].replace('[ZONE]', domain_name)

a1d3s:
> on file powerdns-admin/powerdnsadmin/models/record.py line 178 add on new blank line and
> record["record_data"] = record['record_data'].replace('[ZONE]', domain_name)

Works great

sshipway:
@RobertIchim Can you submit this to the project as a pull request?  Looks like a useful feature that is easy to implement but would need a little update to the documentation to go with it

Woefdram:
Thanks @RobertIchim

[...]

PR #960, opened 2021-06-15
Simple [ZONE] placeholder for "advanced" templating

hvanderheide:
The same solution as proposed in #670 It seems to work quite well and can easily be extended with more placeholders in the future. If this solution is accepted, some additional documentation is needed as well.


mirko:
Any chance this getting merged soon'ish?

Issue #948, opened 2021-06-04
API - Long response time due to GET (all zones) after single PATCH


FunRaiser1:
Hello,

We have an issue with the PowerDNS-Admin API.  

As you can see in the debug log below, we perform a PATCH request to the PowerDNS-Admin API.  
The API command is instantly forwarded to the PowerDNS API.

After this transaction, PowerDNS-Admin performs a full GET for all zones to the PowerDNS-Admin API, which takes 6 seconds to finish.  After these 6 seconds, PowerDNS API is sending a response back to the API client.

Knowing that we are continuously performing many DNS updates using the API, this task is very slow.

Is there any way to avoid that PowerDNS-Admin performs a full GET after a PATCH API command?  We already tried to turn off bg_domain_updates, but that does not seem to help.

Jun  4 13:59:16  gunicorn[2534]: [2021-06-04 13:59:16,710] [helper.py:18] DEBUG - Sending request to powerdns API {'rrsets': [{'name': '13.1.168.192.in-addr.arpa.', 'type': 'PTR', 'ttl': 86400, 'changetype': 'REPLACE', 'records': [{'content': 'michael7.test.eu.'}]}]}
Jun  4 13:59:16  gunicorn[2534]: [2021-06-04 13:59:16,712] [connectionpool.py:226] DEBUG - Starting new HTTP connection (1): 127.0.0.1:8081
Jun  4 13:59:16  pdns_server[1111]: gmysql Connection successful. Connected to database 'powerdns' on '127.0.0.1'.
Jun  4 13:59:16  pdns_server[1111]: [webserver] 7fb24af0-cef1-49f2-8dda-46c9e8393123 127.0.0.1:34448 "PATCH /api/v1/servers/localhost/zones/1.168.192.in-addr.arpa HTTP/1.1" 204 337
Jun  4 13:59:16  gunicorn[2534]: [2021-06-04 13:59:16,735] [connectionpool.py:433] DEBUG - http://127.0.0.1:8081 "PATCH /api/v1/servers/localhost/zones/1.168.192.in-addr.arpa HTTP/1.1" 204 0
Jun  4 13:59:16  gunicorn[2534]: [2021-06-04 13:59:16,777] [domain.py:119] INFO - Found 1972 domains in PowerDNS-Admin
[...]

PR #962, opened 2021-06-18
respect_bg_domain_updates in routes/api

steschuser:
Attemped fix for #948
Have the API call respect the bg_domain_updates switch

### DIFF
--- a/powerdnsadmin/routes/api.py
+++ b/powerdnsadmin/routes/api.py
@@ -973,10 +973,9 @@ def api_zone_subpath_forward(server_id, zone_id, subpath):
 @apikey_can_access_domain
 def api_zone_forward(server_id, zone_id):
     resp = helper.forward_request()
-    if not Setting.get('bg_domain_updates'):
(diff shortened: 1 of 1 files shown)


powerdnsadmin/routes/api.py
```diff
@@ -973,8 +973,10 @@ def api_zone_subpath_forward(server_id, zone_id, subpath):
 @apikey_can_access_domain
 def api_zone_forward(server_id, zone_id):
     resp = helper.forward_request()
-    domain = Domain()
-    domain.update()
+    if not Setting.get('bg_domain_updates'):
+        domain = Domain()
+        domain.update()
+    status = resp.status_code
```
RoeiGanor, line 979: Duplicated status variable

RoeiGanor (changes requested):

powerdnsadmin/routes/api.py
steschuser: Fixed all the things

PR #974, opened 2021-07-08
Unauthorized Access Page

dimpapac:
The eligibility of a user to access PDA, would be better not rely only on the authentication result. Instead an early check that the user has at least one active role on PDA seems to be a common practice. 

This PR implements a feature that redirects authenticated users with no active role on PDA (ie users that do not have any domain or account linked with their username) to a new template page, where they are informed they are not authorized to access this service. Additional information may be included there by the site admin.


steschuser:
Like the get_user_domain method of User model. 

Not sure about the text of the error page, tho

dimpapac:
Ok we changed the text, thanks for the feedback @steschuser

KostasMparmparousis:
Minor update regarding the latest commit:

An objective of ours was to address an issue where users authenticated by some external providers were not redirected to the login page upon clicking the return link, when encountering the Template.

Furthermore, we factored in the "Allow users to create Domain/view History" features, by disabling the Template when these settings are toggled on, as we believe that they are not in line with the general philosophy of the Unauthorized Access Template.

PR #988, opened 2021-08-12
fix: actually store OIDC logout URL

danimo:
This correctly saves the value, but I have not been able to validate
that it actually properly calls the IDP logout URL with the desired effect.
This is due to other issues with the OIDC feature and my IDP.


danimo:
By "validate that it works" I mean whether the PDA logout actually calls the IdPs logout endpoint.

PR #1027, opened 2021-10-20
dyndns: Respond with HTTP header 'WWW-Authenticate' to unauthed requests

mirko:
The common procedure for HTTP Basic Auth is that a client does /not/
immediately send out credentials via an 'Authorization'-header, but to
wait until the server tells the client to do so - which the server
indicates via the 'WWW-Authenticate'-header.

PowerDNS-Admin (and flask in general), though, abort the whole
communication if no Authorization header was found in the initial
request - resulting in '200 "badauth"'.

While this might work for /some/ HTTP clients - which right away add an
Authorization header crafted from provided credentials (via args or
extracted from given URL), this is /not/ standard and /not/ common.

Hence add the 'WWW-Authenticate'-header for every unauthenticated call
checking for dyndns authorisation.

Note, though, this changes the status code from 200 to 401 in this case,
which - given the explanation why 200 was chosen in the first place -
might cause side effects.

### DIFF
--- a/powerdnsadmin/decorators.py
+++ b/powerdnsadmin/decorators.py
@@ -1,7 +1,7 @@
 import base64
 import binascii
 from functools import wraps
-from flask import g, request, abort, current_app, render_template
+from flask import g, request, abort, current_app, Response
 from flask_login import current_user
 
 from .models import User, ApiKey, Setting, Domain, Setting
@@ -301,7 +301,7 @@ def dyndns_login_required(f):
     @wraps(f)
     def decorated_function(*args, **kwargs):
         if current_user.is_authenticated is False:
-            return render_template('dyndns.html', response='badauth'), 200
+            return Response(headers={'WWW-Authenticate': 'Basic'}, status=401)
         return f(*args, **kwargs)
 
     return decorated_function


jbe-dw:
Hello @mirko 

Sorry for the answer delay, things have changed recently here. Can you give me a link to the discussion that you refer in the note, about the choice to keep a 200 RC ?

Thanks

mirko:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication

RFC7617 is more vague, but definitely defines that at least as a *can*-scheme:
"  Upon receipt of a request for a URI within the protection space that
   lacks credentials, the server can reply with a challenge using the
   401 (Unauthorized) status code ([[RFC7235], Section 3.1](https://datatracker.ietf.org/doc/html/rfc7235#section-3.1)) and the
   WWW-Authenticate header field ([[RFC7235], Section 4.1](https://datatracker.ietf.org/doc/html/rfc7235#section-4.1))."
https://datatracker.ietf.org/doc/html/rfc7617

jbe-dw:
LGTM. We will confirm it tomorrow.

mirko:
Thanks for merging. Can you elaborate on the "things have changed recently here" part?
I'm asking since I have quite a few changesets lying around, but didn't bother opening more PRs as - no offense - the project seemed rather unresponsive.

jbe-dw:
You can read all the pinned issues in the repo, but [this one](https://github.com/PowerDNS-Admin/PowerDNS-Admin/issues/1144) should give you an insight.

Just keep in mind that given our free time, we will review bugfix PR but hold feature request for later if there is no associated PR or if the changes are too many.

Regards

PR #1030, opened 2021-10-24
Env oauth url

benshalev849:
This allows the user to bypass the DB registartion of the oAuth variable and put it in as an environment variable.

### DIFF
--- a/configs/docker_config.py
+++ b/configs/docker_config.py
@@ -5,6 +5,9 @@ SQLALCHEMY_DATABASE_URI = 'sqlite:////data/powerdns-admin.db'
 
 legal_envvars = (
     'SECRET_KEY',
+    'PDNS_API_URL',
+    'PDNS_TOKEN_URL',
+    'PDNS_AUTH_URL',
     'BIND_ADDRESS',
     'PORT',
     'LOG_LEVEL',
--- a/configs/docker_config.py
+++ b/configs/docker_config.py
@@ -7,7 +7,7 @@ legal_envvars = (
     'SECRET_KEY',
     'OIDC_OAUTH_API_URL',
     'OIDC_OAUTH_TOKEN_URL',
-    'PDNS_AUTH_URL',
+    'OIDC_OAUTH_AUTHORIZE_URL',
     'BIND_ADDRESS',
     'PORT',
     'LOG_LEVEL',
--- a/configs/docker_config.py
+++ b/configs/docker_config.py
@@ -6,7 +6,7 @@ SQLALCHEMY_DATABASE_URI = 'sqlite:////data/powerdns-admin.db'
 legal_envvars = (
     'SECRET_KEY',
     'OIDC_OAUTH_API_URL',
-    'PDNS_TOKEN_URL',
+    'OIDC_OAUTH_TOKEN_URL',
     'PDNS_AUTH_URL',
     'BIND_ADDRESS',
     'PORT',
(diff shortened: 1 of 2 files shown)


configs/docker_config.py
```diff
@@ -5,6 +5,9 @@
 
 legal_envvars = (
     'SECRET_KEY',
+    'PDNS_API_URL',
```
RGanor, line 8:
```suggestion
    'OIDC_OAUTH_API_URL',
```

configs/docker_config.py
```diff
@@ -5,6 +5,9 @@
...
+    'PDNS_TOKEN_URL',
```
RGanor, line 9:
```suggestion
    'OIDC_OAUTH_TOKEN_URL',
```

configs/docker_config.py
```diff
@@ -5,6 +5,9 @@
...
+    'PDNS_AUTH_URL',
```
RGanor, line 10:
```suggestion
    'OIDC_OAUTH_AUTHORIZE_URL',
```

powerdnsadmin/models/setting.py
```diff
@@ -264,13 +264,23 @@ def set(self, setting, value):
...
-                ] else result.value
+            if ('PDNS_API_URL' in current_app.config) and (setting == 'oidc_oauth_api_url'):
+                result = current_app.config['PDNS_API_URL']
+                return result
+            elif ('PDNS_AUTH_URL' in current_app.config) and (setting == 'oidc_oauth_authorize_url'):
+                result = current_app.config['PDNS_AUTH_URL']
+                return result
+            if ('PDNS_TOKEN_URL' in current_app.config) and (setting == 'oidc_oauth_token_url'):
+                result = current_app.config['PDNS_TOKEN_URL']
+                return result
```
RGanor, line 275:
```suggestion
            if setting.upper() in current_app.config:
                result = current_app.config[setting.upper()]
            else:
                result = self.query.filter(Setting.name == setting).first()
```

powerdnsadmin/models/setting.py
```diff
@@ -264,13 +264,23 @@ def set(self, setting, value):
...
             else:
```
RGanor, line 276:
```suggestion
```

powerdnsadmin/models/setting.py
```diff
@@ -264,13 +264,23 @@ def set(self, setting, value):
...
-                return self.defaults[setting]
+                result = self.query.filter(Setting.name == setting).first()
+                if result is not None:
+                    return strtobool(result.value) if result.value in [
+                        'True', 'False'
+                    ] else result.value
+                else:
+                    return self.defaults[setting]
```
RGanor, line 283:
```suggestion
            if result is not None:
                result = result.value if hasattr(result,'value')
                return strtobool(result) if result in [
                    'True', 'False'
                ] else result
            else:
                return self.defaults[setting]
```

RGanor (changes requested):
I think this can be more generalized to support more environment variable overriding and not to be specific for OIDC variables.

```python
 if setting.upper() in current_app.config:
                result = current_app.config[setting.upper()]
```

Using this method every variable that will be added to `legal_envars` can override database settings.

----
```python
  result = result.value if hasattr(result,'value')
```
This change was needed because we introduced values that does not come from the database `query` result, and we want to preserve `strtobool` convention in the returned value (without code duplication)

Issue #1014, opened 2021-10-08
Salt reuse for API key?


wdoekes:
In https://github.com/ngoduykhanh/PowerDNS-Admin/blob/993e02b635ed271f855e02aeab3e2f8c18d0554f/powerdnsadmin/models/api_key.py#L90-L91 it looks like a predefined SALT is reused for all generated API keys.

Shouldn't that use `bcrypt.gensalt()` instead?

RGanor:
Thank you for bringing this to my attention!
Indeed it seems like this is the right thing to do. Im researching about this topic more and i'll create a fix for it soon.

RGanor:
The point of salt is to create different hash for similar values:
https://en.wikipedia.org/wiki/Salt_(cryptography)#Example_usage

Because API Keys are unique, their hash values are already different, so having the same salt is not a problem.

wdoekes:
I know what a salt is for ;) The point is _not_ -- as you say -- to create a different hash for the same value. The point is to make precomputing hashes harder/impossible. If I know the salt beforehand, I can start precomputing now.

Using a fixed salt is far better than using no salt at all. But here I see no reason here why the salt would have to be fixed/predefined.

zoeller-freinet:
I was going to suggest adding a comment explaining why this kind of usage is OK in this case -
on the other hand using `bcrypt.gensalt()` wouldn't hurt either and would help avoiding similar confusion in the future.

The value of `current_app.config['SALT']` is also used by ItsDangerous to sign tokens for
    email address verification if the `signup_enabled` setting is True. [[1]](https://github.com/ngoduykhanh/PowerDNS-Admin/blob/v0.2.4/powerdnsadmin/services/token.py#L7) By default it's a bcrypt-formatted string as output by `bcrypt.gensalt()`. This is a bit strange as the `salt` parameter serves a different purpose there (see [ItsDangerous Documentation (1.1.x)](https://readthedocs.org/projects/itsdangerous/downloads/pdf/1.1.x/), section 3.2.1 "The Salt"), but it doesn't look like that should cause trouble currently.

> The point is _not_ [...] to create a different hash for the same value.

Making different instances of the same password result in different hashes is of course among the purposes of a salt as used in the context of password hashing.

wdoekes:
> making different instances of the same password result in different hashes is of course among the purposes of a salt as used in the context of password hashing.

[...]

PR #1037, opened 2021-11-08
Clarify salt re-use for API keys

zoeller-freinet:
See https://github.com/ngoduykhanh/PowerDNS-Admin/issues/1014.


wdoekes:
Rather wordy, but far better than without any comments :+1:

PR #1044, opened 2021-11-16
feat: Associate an API Key with accounts

jbe-dw:
As discussed, the feature is added with few changes:

- An API Key can be linked to accounts, as for domains.
- Key is allowed on all domains owned by the linked accounts
- Key with User Role should at least have one domain or one account
- Key with Administrator or Operator role don't have domains or accounts because they are granted anyway. Any update to existing keys will remove all domains.
- The API output has slightly been changed: ApiKeys have an "accounts" key and Accounts have an "apikeys" key

@RGanor The code is functionnal. I tested use cases on my docker instance and it should be fine. But, as I may have told, I don't have much time to spend on the feature so It would be cool if folks could test it and report bugs.

Any comment regarding the adopted behaviour is welcome, and any help looking at the swagger definition would be appreciated.

I open this PR as a preview accessible for PDA Users and contributors but consider it "not finished"


RGanor:
Hey @jbe-dw, Great job with this pull request! this feature was requested for a long time now.
Im testing this right now and I notice something that can cause a lot of confusion. When adding domains/accounts into non user key, then the values are not saved and after refresh the list return to blank.
I understand that this is not a bug, when a key is administrator or operator then it is useless to update its accounts/domains, this is just a cosmetic thing that could lead to confusion.

jbe-dw:
Yes you are right. I was thinking about some js to disable and clear the lists whenever changing the role in the selection box but it's not my cup of tea.

If you are not "more comfortable than me" with the jquery stuff and angulaer things, i'll try to add this in the weekend, but if it is easy for you, you're welcome :smiley_cat: .

jbe-dw:
Also not that i've slightly updated the template locally but not pushed. I have put the account section first and changed the title and content too. I will add a new commit depending on your ability to help me or not on the js stuff above.

jbe-dw:
Added Javascript to enhance behaviour on the ApiKey Edit/Add Page:
- Selecting Administrator or Operator role hide the selection lists, and changing the role to one of these roles clears the selection
- A popup appears if you use one of these roles and you have account or domain selected to warn the user that it won't be saved (though it should not happen because of the previous point).
- A popup appears if you choose a User role and set no account and no domain. The form is not submitted.

Let me know if I have to change something !

Regards

jbe-dw:
There may be a better way to change the selection adressing the widget directly rather than the dom elements, if someone knows ..

RGanor:
Hey @jbe-dw,
Please resolve the new conflicts and I'll merge this ASAP.
Notice that the previous merge have upgraded the schema so I think you'll need to upgrade as well.

jbe-dw:
Hello,

I've updated the swagger file to reflect all the possible error code on the API. It needs a bit of polishing to customize the error messages.

I use it to build my terraform provider, it's cleaner to have a real message rather than "return code is not handled by the client" or a poor "message=({"Error": "", "Errors": []}).

I'll check the conflicts too.

jbe-dw:
I don't get why there was a conflict because I didn't make any change on the conflicting section, but I have rebased and dropped a fix into a separate PR. I'll finish updating the Swagger file tomorrow and will push it afterward.

jbe-dw:
Finally there's too much extra work because return codes and message are unconsistent. I'll open another PR to review all error codes on the API side and change all "abort(HTTP_CODE)" to a proper "raise ERROR" structure with a clear message, including the requested changes, and edit the swagger file accordingly.

jbe-dw:
I let you take a look at the doc changes and check the edit page again. 

Changing a key with user to admin clears the domains and accounts, so changing back to user don't set the previous values again, user has to cancel. 

If LGTM is ok it's ready to merge then.

RGanor:
@jbe-dw have you finished with this pr? I want to test and merge it.

jbe-dw:
Pretty soon. My editor changed tabs for spaces on files or section not concerned by this patch. I reverted it because it was not easy to see what has really changed or not. Doing so and rebuilding gave me a surprise that I have fixed. I want to test it again today before merging.

We introduced broken new features in the past, I want to be sure it is ok. I'll do my best to finish today.

jbe-dw:
Looks OK to me

Issue #1046, opened 2021-11-18
Question: Why accounts name character set is limited


jbe-dw:
Hello,

Do anyone know why we are limited to alphanumeric characters ? It is inconvenient in our use case as we want to add prefixes and avec a naming scheme like:

<env>_<ownerteam>

It is possible to have extra chars like dashes or underscores in the account name in pdns itself.

PR #1002, opened 2021-09-13
Provision PDA user privileges with OIDC Attributes

dimpapac:
As we already mentioned in an earlier stage, provisioning Roles & Associations using a attribute in the user's object is a technique that can be leveraged across multiple Authentication/Authorization providers for PDA.  Continuing on from the last pull request (https://github.com/ngoduykhanh/PowerDNS-Admin/pull/980) we have implemented the OIDC counterpart, including some minor additions to the original documentation too.

PR #1047, opened 2021-11-18
feat: Allow underscores and hyphens in account name

jbe-dw:
#1046 
Allow hyphens and underscores in account names

### DIFF
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -401,7 +401,7 @@ def login():
             if name_prop in me and desc_prop in me:
                 accounts_name_prop = [me[name_prop]] if type(me[name_prop]) is not list else me[name_prop]
                 accounts_desc_prop = [me[desc_prop]] if type(me[desc_prop]) is not list else me[desc_prop]
-		
+
                 #Run on all groups the user is in by the index num.
                 for i in range(len(accounts_name_prop)):
                     description = ''
@@ -411,7 +411,7 @@ def login():
 
                     account_to_add.append(account)
                 user_accounts = user.get_accounts()
-                
+
 		# Add accounts
                 for account in account_to_add:
                     if account not in user_accounts:
@@ -1092,8 +1092,12 @@ def create_group_to_account_mapping():
 
 
 def handle_account(account_name, account_description=""):
+    if Setting().get('account_name_extra_chars'):
+        char_list = "abcdefghijklmnopqrstuvwxyz0123456789_-."
+    else:
+        char_list = "abcdefghijklmnopqrstuvwxyz0123456789"
     clean_name = ''.join(c for c in account_name.lower()
-                         if c in "abcdefghijklmnopqrstuvwxyz0123456789_-")
+                         if c in char_list)
     if len(clean_name) > Account.name.type.length:
         current_app.logger.error(
             "Account name {0} too long. Truncated.".format(clean_name))
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -323,8 +323,8 @@ def login():
                             # Regexp didn't match, continue to next iteration
                             continue
 
-                    account = Account()
-                    account_id = account.get_id_by_name(account_name=group_name)
+                    sanitized_group_name = Account.sanitize_name(group_name)
+                    account_id = account.get_id_by_name(account_name=sanitized_group_name)
 
                     if account_id:
                         account = Account.query.get(account_id)
@@ -345,10 +345,12 @@ def login():
                             current_app.logger.info('User {} added to Account {}'.format(
                                 user.username, account.name))
                     else:
-                        account.name = group_name
-                        account.description = group_description
-                        account.contact = ''
-                        account.mail = ''
+                        account = Account(
+                            name=sanitized_group_name,
+                            description=group_description,
+                            contact='',
+                            mail=''
+                        )
                         account.create_account()
                         history = History(msg='Create account {0}'.format(
                             account.name),
@@ -1092,18 +1094,10 @@ def create_group_to_account_mapping():
 
 
 def handle_account(account_name, account_description=""):
-    if Setting().get('account_name_extra_chars'):
-        char_list = "abcdefghijklmnopqrstuvwxyz0123456789_-."
-    else:
-        char_list = "abcdefghijklmnopqrstuvwxyz0123456789"
-    clean_name = ''.join(c for c in account_name.lower()
-                         if c in char_list)
-    if len(clean_name) > Account.name.type.length:
-        current_app.logger.error(
-            "Account name {0} too long. Truncated.".format(clean_name))
+    clean_name = Account.sanitize_name(account_name)
     account = Account.query.filter_by(name=clean_name).first()
     if not account:
-        account = Account(name=clean_name.lower(),
+        account = Account(name=clean_name,
                           description=account_description,
                           contact='',
                           mail='')
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -331,7 +331,7 @@ def login():
                         # check if user has permissions
                         account_users = account.get_user()
                         current_app.logger.info('Group: {} Users: {}'.format(
-                            group_name, 
+                            group_name,
                             account_users))
                         if user.id in account_users:
                             current_app.logger.info('User id {} is already in account {}'.format(
@@ -351,7 +351,7 @@ def login():
                         account.mail = ''
                         account.create_account()
                         history = History(msg='Create account {0}'.format(
-                            account.name), 
+                            account.name),
                             created_by='System')
                         history.add()
 
@@ -485,13 +485,13 @@ def login():
                                        saml_enabled=SAML_ENABLED,
                                        error='Token required')
 
-        if Setting().get('autoprovisioning') and auth_method!='LOCAL': 
+        if Setting().get('autoprovisioning') and auth_method!='LOCAL':
             urn_value=Setting().get('urn_value')
             Entitlements=user.read_entitlements(Setting().get('autoprovisioning_attribute'))
             if len(Entitlements)==0 and Setting().get('purge'):
                 user.set_role("User")
                 user.revoke_privilege(True)
-                
+
             elif len(Entitlements)!=0:
                 if checkForPDAEntries(Entitlements, urn_value):
                     user.updateUser(Entitlements)
@@ -1093,7 +1093,7 @@ def create_group_to_account_mapping():
 
 def handle_account(account_name, account_description=""):
     clean_name = ''.join(c for c in account_name.lower()
-                         if c in "abcdefghijklmnopqrstuvwxyz0123456789")
+                         if c in "abcdefghijklmnopqrstuvwxyz0123456789_-")
     if len(clean_name) > Account.name.type.length:
         current_app.logger.error(
             "Account name {0} too long. Truncated.".format(clean_name))
--- a/powerdnsadmin/lib/errors.py
+++ b/powerdnsadmin/lib/errors.py
@@ -136,6 +136,13 @@ class AccountNotExists(StructuredException):
         self.message = message
         self.name = name
 
+class InvalidAccountNameException(StructuredException):
+    status_code = 400
+
+    def __init__(self, name=None, message="The account name is invalid"):
+        StructuredException.__init__(self)
+        self.message = message
+        self.name = name
 
 class UserCreateFail(StructuredException):
     status_code = 500
@@ -144,7 +151,6 @@ class UserCreateFail(StructuredException):
         StructuredException.__init__(self)
         self.message = message
         self.name = name
-
 
 class UserCreateDuplicate(StructuredException):
     status_code = 409
--- a/powerdnsadmin/models/account.py
+++ b/powerdnsadmin/models/account.py
@@ -35,7 +35,7 @@ class Account(db.Model):
 
         if self.name is not None:
             self.name = ''.join(c for c in self.name.lower()
-                                if c in "abcdefghijklmnopqrstuvwxyz0123456789")
+                                if c in "abcdefghijklmnopqrstuvwxyz0123456789_-")
 
     def __repr__(self):
         return '<Account {0}r>'.format(self.name)
--- a/powerdnsadmin/models/account.py
+++ b/powerdnsadmin/models/account.py
@@ -34,8 +34,12 @@ class Account(db.Model):
         self.API_EXTENDED_URL = utils.pdns_api_extended_uri(self.PDNS_VERSION)
 
         if self.name is not None:
+            if Setting().get('account_name_extra_chars'):
+                char_list = "abcdefghijklmnopqrstuvwxyz0123456789_-."
+            else:
+                char_list = "abcdefghijklmnopqrstuvwxyz0123456789"
             self.name = ''.join(c for c in self.name.lower()
-                                if c in "abcdefghijklmnopqrstuvwxyz0123456789_-")
+                                if c in char_list)
 
     def __repr__(self):
         return '<Account {0}r>'.format(self.name)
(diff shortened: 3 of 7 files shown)


RGanor:
I think this change will break account auto provisioning in OIDC.
Have you test compatibility?

jbe-dw:
I don't use OIDC. I can get a look at the code or we can summon the folks you did recent updates on this.

jbe-dw:
@benshalev849 @dimpapac 

I can see you worked on OIDC, could you give a feedback about this change's impacts ?

jbe-dw:
@RGanor I'd really enjoy if this could be merged soon, because I need to push my terraform code too and changing the name of the objects will destroy everything and recreate. I'd prefer to have the ability to use these extra characters at first.

I will do my best to double check the OIDC related code tomorrow.

(Looking at the light change in this PR and not using OIDC now I can patch my code locally meanwhile but I'd really prefer to use an official release instead)

benshalev849:
Hey, this can break anyone that has been using OIDC groups authentication atm.

This is because now if the account OIDC created is named:

example_account

the actual account name will be:
exampleaccount

If this change will be pushed new accounts will be created  thus needing to add new permissions.

My suggestions is to add your change as a basic option that the user can toggle.
"Allow underscore in account name"
So it's on the users to choose if they want it or instead of doing it globally for everyone.

jbe-dw:
I looked at the code and updated the corresponding check on the OIDC processing section. What is involved here:

- in the edit page, when submiting the account object formats the name at init, if the entered name does not match the field is marked invalid
- in oidc group import, the check is done the same way in the handle_account function (updated)
- in azure, it seems inconsistent to me. The code looks for a setting in Azure config. The setting is a regex entered by the admin. If it does not match, the account is not created but: if the setting is empty or it passes the regex, the create function is called. The created account is formatted according the pattern in the object model, meaning that the name provided by azure won't match the name in PDA (if I let the setting empty or set it to .*, a group named group+something will be created with the name groupsomething, If the user is in group%something, it will also be added to the same group).
- in the API, no check is done either. You post an account named "group%name" and the API will create and returns "groupname".

The limitation may have been made with the idea that through the API and HTTP URIs we want to limit to characters that do not need encoding (but hyphens, underscore or even the dot don't). It's just a name in a table, so i think it's ok to allow these characters.

I can push a little farther to implement a control in the API and Azure if you want. It should be fast to write the code.

Regards

jbe-dw:
@benshalev849 yes you're right, it logs an error but changes the name. Adding such a setting can be an option.

jbe-dw:
Setting added, defaults to False

jbe-dw:
@zoeller-freinet @RGanor Thank you for your feedback on this one

KostasMparmparousis:
Hi there!
My colleague @dimpapac is unfortunately absent at the moment, so I will respond on his behalf, since I was also a contributor on his OIDC-related PR. Your original change would indeed raise an issue to the Account auto-creating aspect of OIDC Provisioning, in its current format.

However his [PR](https://github.com/ngoduykhanh/PowerDNS-Admin/pull/1002) is basically an alternative to that form of OIDC Provisioning, where one can not only provision a user's associations with Accounts, but also with Domains, as well as the user's role (Administrator, Operator, User) in the Application.

His provisioning has also been implemented and merged for projects that make use of LDAP Authentication, and on top of that your change in its first manifestation would be fully functional if this provisioning were to be used.

jbe-dw:
@KostasMparmparousis Thanks for the feedback. I'm not quite sure about what you are talking about. I guess my english lacks a bit here so my answer here may not match.

First, this PR is about the Account name. I looked at @dimpapac PR (#1002) which seems to be about Users, which is not in the scope of this one. I've reviewed the OIDC current code and, though i'm not using it here, I could see what @benshalev849 's remark was about: 

The automatic account provisioning just cleans out unauthorized characters from the created account name. There is no 1<>1 binding and we can run into issues like "account_A" and "account-A" and any combination of "ac-count.a" would result in the creation of the "accounta" account for all these three cases.

Although this patch wouldn't really solve it, a raw change in the allowed character would result in changing the former situation to the creation of 3 distinct accounts and messing up with ownership of domains and user's membership.

By adding a toggle option to enable this (default disabled), it should have 0 impact on existing content, unless unwisely enabled by an admin. What is your opinion ?

KostasMparmparousis:
Your confusion is understandable, as the title only refers to users, however [PR #1002 ](https://github.com/ngoduykhanh/PowerDNS-Admin/pull/1002) also features Domain and Account Association Autoprovisioning for Users, expanding on the current form of *User-Account* autoprovisioning in PowerDNS-Admin, as implemented below:
```
 #Gets the name_property and description_property.
            name_prop = Setting().get('oidc_oauth_account_name_property')
            desc_prop = Setting().get('oidc_oauth_account_description_property')
            account_to_add = []
	    #If the name_property and desc_property exist in me (A variable that contains all the userinfo from the IdP).
            if name_prop in me and desc_prop in me:
                accounts_name_prop = [me[name_prop]] if type(me[name_prop]) is not list else me[name_prop]
                accounts_desc_prop = [me[desc_prop]] if type(me[desc_prop]) is not list else me[desc_prop]
		
                #Run on all groups the user is in by the index num.
                for i in range(len(accounts_name_prop)):
                    description = ''
                    if i < len(accounts_desc_prop):
                        description = accounts_desc_prop[i]
                    account = handle_account(accounts_name_prop[i], description)
                    account_to_add.append(account)
                user_accounts = user.get_accounts()
		# Add accounts
                for account in account_to_add:
                    if account not in user_accounts:
                        account.add_user(user)
```
Although @benshalev849 remark is indeed correct,  our suggestion was that, by using @dimpapac 's Autoprovisioning, account ownership could remain unaffected without the need to modify your contribution's original implementation.

Concerning your feature now, a toggle switch would be the only option in order to maintain compatibility for projects that use the current autoprovisioning method.

powerdnsadmin/routes/index.py
```diff
@@ -1056,8 +1056,12 @@ def create_group_to_account_mapping():
 
 
 def handle_account(account_name, account_description=""):
+    if Setting().get('account_name_extra_chars'):
```
zoeller-freinet, line 1059: Could you please place the account name sanitization logic into a dedicated function which you call from within both `models.account.Account.__init__` and `routes.index.handle_account`? E.g. roughly:

```python3
class Account(db.Model)
    # ...
    def __init__(self, name=None, description=None, contact=None, mail=None):
        self.name = Account.sanitize_name(name) if name is not None else name
        # ...
    # ...

    
    @staticmethod
    def sanitize_name(name):
        allowed_characters = "abcdefghijklmnopqrstuvwxyz0123456789"
        if Setting().get('account_name_extra_chars'):
            allowed_characters += "_-."
        sanitized_name = ''.join(c for c in name.lower() if c in allowed_characters)
        if not sanitized_name:
            raise InvalidAccountNameException("Empty string is not a valid account name")
        return sanitized_name
```
That should make intent clearer and also unifies the logic into one sensible place, improving maintainability.

powerdnsadmin/routes/index.py
```diff
@@ -1056,8 +1056,12 @@ def create_group_to_account_mapping():
...
+        char_list = "abcdefghijklmnopqrstuvwxyz0123456789_-."
+    else:
+        char_list = "abcdefghijklmnopqrstuvwxyz0123456789"
     clean_name = ''.join(c for c in account_name.lower()
-                         if c in "abcdefghijklmnopqrstuvwxyz0123456789")
+                         if c in char_list)
     if len(clean_name) > Account.name.type.length:
```
zoeller-freinet, line 1065: > in oidc group import, the check is done the same way in the handle_account function (updated). Tn hough it logs an error "Name truncated" it's still created

This actually checks against max length of the string column in the DB model, so it doesn't have much to do with this PR, but it nevertheless should be moved into `models.account.Account` on this occasion as well.

zoeller-freinet (changes requested):

zoeller-freinet:
@jbe-dw
I added two comments on the code, could you please take a look? Please also resolve the merge conflicts.

---

Trying to summarize:
- Not affected at this time as there is currently no feature that auto-provisions accounts via these mechanisms:
  - Google OAuth
  - GitHub OAuth
- Will be handled by `Account.sanitize_name`:
  - API
  - Microsoft Azure OAuth
  - OpenID Connect

For LDAP, account *associations* can already be auto-provisioned if the `autoprovisioning` Setting is enabled ("Roles Autoprovisioning" radio button "ON" on the LDAP auth tab) [(code entry point)](https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/328780e2d41b7120afecf22c83b8d88864eaeaf7/powerdnsadmin/routes/index.py#L490). In this case, the account (name) has to exist in PDA's DB beforehand - the purported account name is used as-is for this lookup. If it does not exist, it is not created and the user is consequently not associated with the purported account. This is the mechanism that PR #1002 seeks to make usable for OIDC as well.

Please mention if something is incorrect or missing.

@KostasMparmparousis 
I don't understand how PR #1002 would remove the need to add a default-off compatibility setting that allows switching additional allowed characters for account names on or off. The PR does not cover the API and Azure OAuth cases. As for OIDC, supposing that #1002 is merged in its current form, there would be two ways to autoprovision accounts via OIDC: the one that is already part of `master` (the code you quoted), and the one facilitated by #1002. Users of the former would still require a toggle switch for compatibility, wouldn't they?

I haven't really looked into #1002 yet, and would prefer to consider it separate from this PR here. In general, adding compatibility toggle switches should be avoided if possible, and I'm not that keen on doing it, but it seems like a pragmatic solution in this case. I imagine it should be possible to remove the switch again and expand the set of allowed characters by default without losing account associations by triggering an interactive "migration" on SSO login as a way out. Maybe something along those lines could be implemented sometime in the future.

powerdnsadmin/routes/index.py
jbe-dw: Hi,

I don't agree with your comment, as `clean_name` is the formatted name from the character_list, and is used later to check/create the account.

Look [here](https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/23dcef764c37f17eb8ba3e21b4d924832ffa4efc/powerdnsadmin/routes/index.py#L1068-L1074) for the snippet. This is why I updated the code here as well. This would change the creation because the search would return no match with the new extended name, leading in a new account creation/association, but without updating the zones ownership.

Anyway, I'll use the model method instead.

powerdnsadmin/routes/index.py
jbe-dw: The amazing thing with you is that you provide the code you asked to change !

jbe-dw:
This could have been avoided if the application just rejected names not matching the allowed characters. This is also problematic on a terraform side. Terraform compares what user request and what leaves in the API. It means that if a user provides a name with unallowed character, the API would accept to create it and silently change it. On the next run, Terraform would see that the name with the id it received upon creation is different, and would try to update it, with no real effect in PDA. The code will then always request an impossible change, without failing.

This mean that it should be better to change the behaviour to reject the names that do not match the allowed character, in all the concerned modules (GUI/API or autoprovisioning).

@zoeller-freinet concerning Azure, as I said, the name is matched again a regex that is a setting. Calling the sanitized name would break the logic. It may need a bit of work. I don't remind it well (i was on holidays) I'll have a look shortly.

I'll also review the code for both OIDC and LDAP, to be sure.

powerdnsadmin/routes/index.py
zoeller-freinet: Sorry if I was unclear, I was only referring to [these](https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/23dcef764c37f17eb8ba3e21b4d924832ffa4efc/powerdnsadmin/routes/index.py#L1065-L1067) three lines, which lend themselves to being moved into the `Account` class as well on this occasion. Emitting an error log line if the account name (sanitized or not) will be truncated before written into the DB because it doesn't fit into max column character length shouldn't in itself have much to do with this PR.

zoeller-freinet:
> This could have been avoided if the application just rejected names not matching the allowed characters. This is also problematic on a terraform side. Terraform compares what user request and what leaves in the API. It means that if a user provides a name with unallowed character, the API would accept to create it and silently change it. On the next run, Terraform would see that the name with the id it received upon creation is different, and would try to update it, with no real effect in PDA. The code will then always request an impossible change, without failing.
>
> This mean that it should be better to change the behaviour to reject the names that do not match the allowed character, in all the concerned modules (GUI/API or autoprovisioning).

That's right, and this should be done. But it would not mitigate the need to add a compatibility setting like `account_name_extra_chars`, since we should assume that there already exist PDA DB instances that contain account names that originated from Azure or OIDC, and were then bent into a different account name by PDA before storage into PDA's DB.

> concerning Azure, as I said, the name is matched again a regex that is a setting. Calling the sanitized name would break the logic. It may need a bit of work. I don't remind it well (i was on holidays) I'll have a look shortly.

Based on your comment on Azure [here](https://github.com/PowerDNS-Admin/PowerDNS-Admin/pull/1047#issuecomment-983535689) and my study of the code, I didn't notice any *additional* issue based on the state of this PR right now that isn't yet handled, but I didn't use Azure OAuth so far and may be missing something. I'll await your investigation.

powerdnsadmin/routes/index.py
jbe-dw: After looking at the code, I have more questions. There is something I don't get. I moved the sanitize function in the Account model, with the length check as requested, which just logs an error. Running Account.query.filter(account_name == something_over_40_chars) like it is done should return no value, triggering a creation that would fail because the name would be duplicate ? Or the filter is intelligent enough to truncate the requested name to match the field length ?

I could also see that in several places, when the function checks if the Account exists, it's done with the provided name. Either a call to  Account.query.filter sanitizes the name, which I expect it does not, and it's fine, or it does not and the check is always negative, but the account creation should then fail.

You can answer that I can run test to see how it behaves, but another pov is welcome.

The underlying question is: should we truncate the name in the sanitize function either, and filter by the sanitized name instead ?

KostasMparmparousis:
Hello there!

> For LDAP, account _associations_ can already be auto-provisioned if the `autoprovisioning` Setting is enabled ("Roles Autoprovisioning" radio button "ON" on the LDAP auth tab) [(code entry point)](https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/328780e2d41b7120afecf22c83b8d88864eaeaf7/powerdnsadmin/routes/index.py#L490). In this case, the account (name) has to exist in PDA's DB beforehand - the purported account name is used as-is for this lookup. If it does not exist, it is not created and the user is consequently not associated with the purported account. This is the mechanism that PR #1002 seeks to make usable for OIDC as well.

First of all, i would like to comment that you are spot on in regards to the functionality of LDAP Autoprovisioning and the rationale behind PR #1002.

> @KostasMparmparousis I don't understand how PR #1002 would remove the need to add a default-off compatibility setting that allows switching additional allowed characters for account names on or off. The PR does not cover the API and Azure OAuth cases.

Furthermore, you are indeed correct. PR #1002 does not cover the API and Azure OAuth cases, only OpenID Connect. However, we felt the need to jump in,  in order to clarify that as an alternative to the current implementation, PR #1002 would not be affected by the proposed feature.

> As for OIDC, supposing that #1002 is merged in its current form, there would be two ways to autoprovision accounts via OIDC: the one that is already part of `master` (the code you quoted), and the one facilitated by #1002. Users of the former would still require a toggle switch for compatibility, wouldn't they?
 
Finally, your intuition is correct. The two features do need to be mutually exclusive from one another. With that in mind, when enabling OIDC autoprovisioning ("Roles Autoprovisioning" radio button "ON"), we disable the current form of OIDC Account Autoprovisioning, upon confirmation from the user who gets shown this modal.

Screenshot_1

powerdnsadmin/routes/index.py
jbe-dw: The test showed what I expected: The check don't match but the creation fails with a duplicate. 

I've updated the code to force the character limit in the sanitize function and the checks to compare the sanitized version, in both API, OIDC and Azure auto-provisioning.

Let me know if it suits your expectations and if the syntax style is good to you.

jbe-dw: Concerning LDAP, I'm not sure if we must enforce the sanitized name of the account existence check.

Issue #1010, opened 2021-09-30
Feature: Enable OTP on first Login for all User


a1d3s:
Hello,
how can i force OTP to all Users on first login?
So the User havn't to enable it by himself.
Old One: https://github.com/ngoduykhanh/PowerDNS-Admin/issues/550

a1d3s:
@ngoduykhanh is there any option?

a1d3s:
That it may doesn't exist is not what I ask. I need this as feature. 
Why should I close the issue? Maybe someone can add this option in the backend?

seraphyn:
Cu
> Why should I close the issue?
Had I Said that?
😉

Stay safe and healthy

zoeller-freinet:
This should be implemented now by #1051.

PR #1051, opened 2021-11-22
Enable OTP on first Login

vmarkop:
#1010

This PR forces all users, including existing ones and regardless of their Role, to use OTP, if the respective Setting is toggled, and helps them by introducing a new template that shows their QR code on the first login after the setting is enabled.

Developed by [our team](https://github.com/gunet) in collaboration with @kkmanos


a1d3s:
Works well, thank you

powerdnsadmin/routes/index.py
```diff
@@ -170,6 +172,9 @@ def login():
         login_user(user, remember=False)
         session['authentication_type'] = 'OAuth'
         signin_history(user.username, 'Google OAuth', True)
+        if Setting().get('otp_first_login') and Setting().get('otp_field_enabled') and (user.otp_secret == "" or user.otp_secret is None):
```
zoeller-freinet, line 175: Can you put the multiple occurrences of these three lines into a new function that you call from multiple places and use a single block comment inside it?

You can simplify `(user.otp_secret == "" or user.otp_secret is None)` to `not user.otp_secret`.

zoeller-freinet:
Hi @vmarkop,

thanks a lot for implementing this feature! I'll test it soon.

1. Would e.g. `otp_force` be a better name for this setting? Currently it's called `otp_first_login`, but it forces OTP usage not only on first login, but always - it's possible to disable OTP in the user profile and then the policy will be re-enforced on next login.
2. Would it be possible to make the user enter the OTP on `register_otp.html` before it's possible to continue? This would ensure that the seed has been scanned correctly, preventing a scenario where the user would be locked out later.

vmarkop:
Thank you for your feedback, @zoeller-freinet ! The PR has been updated, users now have to confirm the OTP before they can continue.

vmarkop:
Updated the PR to handle issue #1063.

powerdnsadmin/templates/user_profile.html
```diff
@@ -152,5 +160,15 @@ <h3 class="box-title">Edit my profile{%  if session['authentication_type'] != 'L
         };
         applyChanges(postdata, $SCRIPT_ROOT + '/user/profile', false, true);
     });
+
+    // copy otp secret code to clipboard
+    function copy_to_clickboard() {
```
zoeller-freinet, line 165: Could you rename that function to e.g. `copy_otp_secret_to_clipboard()`, move it to `static/custom/js/custom.js` and delete the duplicate inside `register_otp.html`? 

In `user_profile.html`, `custom.js` is already loaded, and for `register_otp.html`, it would be possible to to add `custom.js` to the `js_login` asset bundle defined in `assets.py`.

powerdnsadmin/routes/index.py
```diff
@@ -684,6 +700,31 @@ def register():
...
 
 
+# Show welcome page on first login if otp_force is enabled
+@index_bp.route('/welcome', methods=['GET', 'POST'])
+def welcome():
+    if 'welcome' not in session:
+        return redirect(url_for('index.index'))
+
+    user_id = session['user_id']
+    user = User(id = user_id)
```
zoeller-freinet, line 710: Please change to `user = User(id=user_id)`.

powerdnsadmin/routes/index.py
```diff
@@ -584,6 +575,24 @@ def get_azure_groups(uri):
...
+        user.update_profile(enable_otp=True)
+        user_id = current_user.id
+        prepare_welcome_user(user_id)
+        return redirect(url_for('index.welcome'))
+    return redirect(url_for('index.login'))
+
+# Prepare user to enter /welcome screen, otherwise they won't have permission to do so
+def prepare_welcome_user(user_id):
+    logout_user()
+    session['user_id'] = user_id
```
zoeller-freinet, line 594: You could rename that dict key to `welcome_user_id`. That would have the advantage of making potential naming conflicts less likely, and you'll also only need 1 key as you'll be able to simply check for existence of the `welcome_user_id` key inside the `welcome()` view function.

powerdnsadmin/routes/index.py
```diff
@@ -684,6 +700,31 @@ def register():
...
+# Show welcome page on first login if otp_force is enabled
+@index_bp.route('/welcome', methods=['GET', 'POST'])
+def welcome():
+    if 'welcome' not in session:
+        return redirect(url_for('index.index'))
+
+    user_id = session['user_id']
+    user = User(id = user_id)
+    login_user(user)
+    encoded_img_data = base64.b64encode(qrcode()[0])
```
zoeller-freinet, line 712: Could you take a look at `routes.user.qrcode()`? There are these three lines of code:
```python3
img = qrc.make(current_user.get_totp_uri(),
               image_factory=qrc_svg.SvgPathImage)
stream = BytesIO()
img.save(stream)
```
Can you move these into a separate function inside `models/user.py` (return `stream.getvalue()`), which you then call from within `routes.user.qrcode()` and in the line marked here? Then you don't have to call `login_user(user)` here just to be able to fetch the qrcode from the view function.

 Doing this would be fine because the user is already authenticated since `welcome_user_id` has been placed into the session after the user was either authenticated or registered.

powerdnsadmin/routes/index.py
```diff
@@ -674,7 +683,14 @@ def register():
                 if result and result['status']:
                     if Setting().get('verify_user_email'):
                         send_account_verification(email)
-                    return redirect(url_for('index.login'))
+                    if Setting().get('otp_force') and Setting().get('otp_field_enabled'):
+                        login_user(user)
```
zoeller-freinet, line 687: This line can be removed, simply replace `current_user` with `user` in the following two lines.

zoeller-freinet (changes requested):
Hi @vmarkop,

great work! I've just reviewed and tested your changes, and left a couple comments in the code - could you take a look? After they have been addressed this should be ready for merge.

powerdnsadmin/routes/index.py
vmarkop: I haven't fully understood this comment.

My solution was to replace lines 709,710 with:
`user = User(id = session['welcome_user_id'])`.

Please let me know if this isn't ideal.

zoeller-freinet: Ah - my fault! Semantically it's all good, just remove the spaces around the = sign for the keyword argument. PEP 8 [recommends this](https://www.python.org/dev/peps/pep-0008/#other-recommendations) in order to visually distinguish keyword arguments from variable assignments.

Could you please also remove [this now unused import](https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/60b84162fa5207b0b48e0d193245a404358bcfdc/powerdnsadmin/routes/index.py#L15)?

vmarkop: Oh ok, I get it now, thanks for the explanation. Done!

zoeller-freinet:
Thank you! Merged.

PR #1053, opened 2021-11-23
Add Keycloak documentation

nickbouwhuis:
I recently implemented PDA auth with Keycloak and thought this documentation might be useful.

### DIFF
--- a/docs/oauth.md
+++ b/docs/oauth.md
@@ -19,6 +19,29 @@ Now you can enable the OAuth in PowerDNS-Admin.
 
 This should allow you to log in using OAuth.
 
+#### Keycloak
+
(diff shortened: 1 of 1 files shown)


RGanor:
Hey @nickbouwhuis , thank you for your contribution!
I see that there is already OIDC documentation, how does keycloak differ from it?

nickbouwhuis:
> Hey @nickbouwhuis , thank you for your contribution! I see that there is already OIDC documentation, how does keycloak differ from it?

There are some specifics and different semantics used in the existing documentation and the web page. For example, the web ui talks about a 'Client key' instead of a 'Client ID'. Also, it isn't really clear that one should use the 'profile' scope. I thought, the more provider documentation we have, the better.

PR #1056, opened 2021-11-28
Docker-compose improvements

turip:
This change ensures that the docker-compose stack is always
restarted when an error occurs or if the host system gets
rebooted.

Also this change allows for setting the database URL via an
environment variable, so that for basic use-cases the docker-compose
file does not need to be edited.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,16 +39,25 @@ $ docker run -d \
 This creates a volume called `pda-data` to persist the SQLite database with the configuration.
 
 #### Option 2: Using docker-compose
-1. Update the configuration   
-   Edit the `docker-compose.yml` file to update the database connection string in `SQLALCHEMY_DATABASE_URI`.
+1. Update the configuration if needed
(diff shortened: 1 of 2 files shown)


jbe-dw:
I used a mix of the docker-compose and test to do latest developments.

It may be improved, to have an intermediate env with both a pdns server and a sqlite db on a persistent volume.

@RGanor ?

AzorianSolutions:
@turip @jbe-dw I recommend you take a look at my [PR for issue 1070 here](https://github.com/ngoduykhanh/PowerDNS-Admin/pull/1072). I have solved all of your concerns and many more.

jbe-dw:
I'll take a look, if I don't have to pick my customization each time I create a test branch it's better, thank you.

turip:
> @turip @jbe-dw I recommend you take a look at my [PR for issue 1070 here](https://github.com/ngoduykhanh/PowerDNS-Admin/pull/1072). I have solved all of your concerns and many more.

Thank you, for pointing out. That one seems way more elaborate, closing down this PR.

Issue #1055, opened 2021-11-24
User can disable DNSSEC


larslehmann-net:
I just found out by accident that a user API key can disable DNSSEC for his domains. In the settings dnssec_admins_only is enabled, so a normal user should not be able to make any changes to the DNSSEC settings.

## How to reproduce
```bash
ZONE=example.net
API_URL=pdns.example.net
API_KEY=XXXXXXXXXXXXX
CRYPTOKEY=$(curl -X GET -H "X-API-Key: ${API_KEY}" -s https://${API_URL}/api/v1/servers/localhost/zones/${ZONE}/cryptokeys | jq '.[0].id')
curl -X DELETE -H "X-API-Key: ${API_KEY}" -s https://${API_URL}/api/v1/servers/localhost/zones/${ZONE}/cryptokeys/${CRYPTOKEY}
```

zoeller-freinet:
This should be fixed now by 07f0d21.

PR #1058, opened 2021-11-30
fix: Check user zone create/delete permission

jbe-dw:
Hello,

I was looking at the API code to update the Swagger definition and I saw that the options allowing a user to create or remove a domain are checked on the /powerdnsadmin/zones endpoint but not on the /servers/localhost/zones endpoint.

These two decorators do the check.

### DIFF
--- a/powerdnsadmin/decorators.py
+++ b/powerdnsadmin/decorators.py
@@ -246,6 +246,46 @@ def api_can_create_domain(f):
     return decorated_function
 
 
+def apikey_can(action, http_methods=[]):
+    """
+    Grant access if:
+        - user is in Operator role or higher, or
+        - the action is allowed for a User in the settings
+    """
+    def decorator(f):
+        @wraps(f)
+        def decorated_function(*args, **kwargs):
+
+            p = {
+                "create_domain": {
+                    "setting": "allow_user_create_domain",
+                    "msg": "create domains",
+                    "toggle": True
+                },
+                "remove_domain": {
+                    "setting": "allow_user_remove_domain",
+                    "msg": "remove domains",
+                    "toggle": True
+                },
+                "dnssec": {
+                    "setting": "dnssec_admins_only",
+                    "msg": "create domains",
+                    "toggle": False
+                }
+            }
+
+            if (g.apikey.role.name not in ['Administrator', 'Operator']
+                and request.method in http_methods
+                and Setting().get(p[action]['setting']) != p[action]['toggle']):
+                msg = "ApiKey #{0} does not have enough privileges to {1}"
+                current_app.logger.error(msg.format(g.apikey.id, p[action]['msg']))
+                raise NotEnoughPrivileges()
+            return f(*args, **kwargs)
+
+        return decorated_function
+    return decorator
+
+
 def apikey_can_create_domain(f):
     """
     Grant access if:
--- a/powerdnsadmin/decorators.py
+++ b/powerdnsadmin/decorators.py
@@ -246,6 +246,45 @@ def api_can_create_domain(f):
     return decorated_function
 
 
+def apikey_can_create_domain(f):
+    """
+    Grant access if:
+        - user is in Operator role or higher, or
+        - allow_user_create_domain is on
+    """
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        if g.apikey.role.name not in [
+                'Administrator', 'Operator'
+        ] and not Setting().get('allow_user_create_domain'):
+            msg = "ApiKey #{0} does not have enough privileges to create domain"
+            current_app.logger.error(msg.format(g.apikey.id))
+            raise NotEnoughPrivileges()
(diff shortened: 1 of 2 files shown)


zoeller-freinet:
Hi @jbe-dw,

thanks for this fix!

For `apikey_can_remove_domain`, instead of defining which HTTP methods to check against inside the decorator, can you implement it as a decorator function that optionally accepts a list of strings (HTTP method names to check against) as an argument? If the list is falsy, permissions are checked regardless of HTTP method. Roughly:

```python3
def apikey_can_remove_domain(only_check_http_methods=[]):
    def decorator(f):
        @wraps(f)
        def decorated_function(*args, **kwargs):
            if not only_check_http_methods or request.method in only_check_http_methods:
                # ... existing checks ...
            return f(*args, **kwargs)
        return decorated_function
    return decorator

@apikey_can_remove_domain(only_check_http_methods=['DELETE'])
def api_zone_forward(server_id, zone_id):
    pass
```

That should decouple the permission checking logic better from the API definition, clarifying intent better and making it easier to reuse the decorator. I believe such a pattern will be helpful for fixing #1055 as well.

jbe-dw:
Hi,

I think it's a good idea however you need to implement different checks for the issue. If so, we also have to check which component should be checked, maybe by adding a second parameter. We would change the function name to something more generic like "apikey_can" and merge the decorator to check the creation.

This way you could reuse the logic to add a check for the cryptokeys (or I can do it meanwhile).

zoeller-freinet:
Yes, sorry - fixing #1055 will require implementation of another decorator as I currently see it. I didn't mean to check permissions for both `apikey_can_remove_domain` and `apikey_can_configure_dnssec` using the same decorator.

For example, all indications are that with `/servers/{server_id}/zones/{zone_id}/cryptokeys`, only `POST` needs an additional check, while with `/servers/{server_id}/zones/{zone_id}/cryptokeys/{cryptokey_id}`, the check must run on all methods, so the pattern will be useful there too.

Implementing even more generic decorators for permission checking is an interesting idea as well that could be considered, but probably out of scope of this PR.

jbe-dw:
I say that because I hesitated to do so while pushing this one, as the two functions are siblings. It could be a first shot at it. I'll do something about it this week, there's not much work.

Regards

jbe-dw:
Please look at the last commit for a generic function

zoeller-freinet:
I see what you mean and it's tempting, but I think it has a tendency of becoming too complex for a couple of reasons. Permission-checking code should be as clear-cut and easy to read as possible.

Independently from that, I believe it would be a lot more natural for permission-checking decorators to accept and act on a list of HTTP methods (`api_bp.route` already does), than basic settings, which would be a larger step conceptually.

Regarding 14092c9 again, also consider that yours would be the first permission-checking decorator that defines inside itself a HTTP method to check against (although that may simply be a result of the need to only check against certain HTTP methods not having been acknowledged as of yet), while the same is already the case for roles and basic settings with many decorators.

My suggestion would be to:
- Either revert to 14092c9 (I'd be OK with merging it)
- Or build upon 14092c9 by implementing something like https://github.com/ngoduykhanh/PowerDNS-Admin/pull/1058#issuecomment-983850788 for `apikey_can_remove_domain`

What do you think?

jbe-dw:
OK I get it. I just provided a base code for a more generic function based on your comment. If you think it's a bad design to mix in the same decorator several checks, I'm OK with it (in my mind I though that several API routes could implement a standard 'apikey_can' function which would be easier than finding the right decorator to add).

I can apply your second suggestion if you think it is better. But I think that if you don't want to have a generic decorator for simplicity, then this decorator will only apply the zone deletion endpoint that is a 'DELETE' http method, so adding a parameter to the decorator that will only be 'DELETE' whatsoever is pointless.

Tell me if you prefer suggestion #2 i can add it, or just drop commit 5d77e5c9

jbe-dw:
I have a busy partying weekend starting right now ! I'll push what you decided on Sunday (or feel free to make the changes if you want to merge before).

zoeller-freinet:
@jbe-dw
> this decorator will only apply the zone deletion endpoint that is a 'DELETE' http method, so adding a parameter to the decorator that will only be 'DELETE' whatsoever is pointless.

I'm not so sure what you mean? For `api_zone_forward`, the permission must not be checked on methods other than `DELETE`, and specifying the method outside of the decorator function body, within the API definition, has in my opinion the mentioned conceptual advantages.

Anyway, I implemented suggestion 2 and merged it, see d2f35a4059fc1455d548e4598ca6067778376c54 (it seems I had no write access to https://github.com/jbe-dw/PowerDNS-Admin/tree/fixAPIUserPermissionCheck). I hope you had a nice party!

jbe-dw:
I think you get what I meant (it is possibly a better design, though I'm not a developer but an ops with dev skills - mainly python to tell otherwise).

Thanks for the update, I had a great party for sure !

PR #1042, opened 2021-11-10
History Tab Overhaul & Domain Record Modifications Changelog

kkmanos:
This PR is a continuation of the raised issue [#1025](https://github.com/ngoduykhanh/PowerDNS-Admin/issues/1025).

We have created a [wiki page](https://github.com/gunet/PowerDNS-Admin/wiki/Domain-Changelog-&-History-Overhaul) providing full documentation on the proposed feature.

This PR also addresses issue [#872](https://github.com/ngoduykhanh/PowerDNS-Admin/issues/872), by adding a zone-specific changelog.

Note that the Domain Changelog does not contain Reverse records that were created automatically with the `auto_ptr` setting enabled, because they do not produce a history log, as stated in issue [#1041](https://github.com/ngoduykhanh/PowerDNS-Admin/issues/1041).

PR #1059, opened 2021-12-01
Datepicker replace

kkmanos:
Replaced jquery-ui-datepicker with bootstrap datepicker. Also removed obsolete static files. This PR is an update of #1042


powerdnsadmin/assets.py
```diff
@@ -58,6 +59,8 @@ def concat(self, out, hunks, **kw):
                  'node_modules/jtimeout/src/jTimeout.js',
                  'node_modules/jquery.quicksearch/src/jquery.quicksearch.js',
                  'custom/js/custom.js',
+                 'node_modules/bootstrap-datepicker/js/bootstrap-datepicker.js',
```
zoeller-freinet, line 62: Could you remove this line? Will merge subsequently.

kkmanos: Thanks for the remark. 
I added the change you suggested.

zoeller-freinet: Thanks! Merged.

PR #1061, opened 2021-12-01
Update base.html

ciscoqid:
Fixed reference to jquery-ui.css from http to https (may not be the proper fix). Using http causes browsers to be angry and resentful.


RGanor:
This issue already solved in the master

Issue #1064, opened 2021-12-02
History crash in Master


jbe-dw:
Hello,

There's a bug in the detailed history patch. Updating zone ownership generates an history with an empty string detail. Calling json.loads throw an exception. The domain dashboard page display a 500 error.

Regards

zoeller-freinet:
I can't seem to reproduce this on 6c1dfd2. I'm using a fresh Ubuntu 20.04 installation, https://github.com/ngoduykhanh/PowerDNS-Admin/wiki/Running-PowerDNS-Admin-with-Systemd,-Gunicorn--and--Nginx and `configs/development.py` as PDA configuration, starting from an empty DB.

Can you describe detailed steps that result in the exception being thrown?

jbe-dw:
If you change the account owning a domain through the API it registers an history message without detail, or delete through the API.

We could update the api to insert a json instead. There's actions not writing to the history (changing owner via the gui does not log the operation).

zoeller-freinet:
Getting the crash now, thanks!

> We could update the api to insert a json instead. There's actions not writing to the history (changing owner via the gui does not log the operation).

If the history logic generally expects the detail column to contain a JSON object, then yes, we should insert `{}` by default if there are no details (I haven't really looked at that part of PDA's code in detail yet). I think it should nevertheless be robust enough to not crash if the detail field is empty.

Given #1066 I think we can close this.

PR #1065, opened 2021-12-02
fix: History crash when undetailed history

jbe-dw:
Fix for #1064


jbe-dw:
Feel free to close this PR if 1066 is merged

PR #1071, opened 2021-12-07
fix: Error in the swagger AccountSummary definition

jbe-dw:
Hello,

It may be a detail but i couldn't generate a client that works with the old definition. It waited for a list of list of dicts because the referenced object is already a list.

Regards


zoeller-freinet:
Looks correct which a test of `GET /pdnsadmin/apikeys` and `GET /pdnsadmin/users/test` confirmed. Thanks for reporting it.

PR #1073, opened 2021-12-08
chore: remove funding and sponsor badges

ngoduykhanh:
As I am not actively maintaining this project anymore.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,0 @@
-github: [ngoduykhanh]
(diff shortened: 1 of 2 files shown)


AzorianSolutions:
@ngoduykhanh If that is the case, would you consider transferring ownership of this project to me for ongoing maintenance?

ngoduykhanh:
@AzorianSolutions Actually we have 2 active maintainers now. I don't think I will transfer the ownership to an individual, perhaps it should be an Organization.

AzorianSolutions:
@ngoduykhanh I am backed by two organizations here, my own and an employer's. I need to ensure that what ever GUi solution I use for managing our production DNS systems, provides a specific desired outcome. That is my value proposition.

Never the less, perhaps I misunderstood your intent? I read the message as though the project would be abandoned by your account.

PR #1076, opened 2021-12-10
Add button for admin page in single Domain view

milefork:
While working with PDA I had often the use to get from the Domain Records to the Admin page of it, to for example assign an account to it.

Therefore it would be nice to have a button for it to get to there.

example

### DIFF
--- a/powerdnsadmin/templates/domain.html
+++ b/powerdnsadmin/templates/domain.html
@@ -25,6 +25,9 @@
                     <button type="button" class="btn btn-flat btn-primary pull-left button_add_record" id="{{ domain.name }}">
                         Add Record&nbsp;<i class="fa fa-plus"></i>
                     </button>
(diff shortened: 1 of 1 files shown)


zoeller-freinet:
Hi @milefork, thank you for your first contribution!

Having that button there makes sense, however it is currently displayed even if the logged-in user does not have permission to administrate domains (resulting in an HTTP 403 response if the button is clicked in this case). Can you update this PR so that the button is only rendered if the user has that permission? You can look at [`dashboard_domain.html`](https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/7808febad8fb0a56cdecce2d7cdb9a92ec30ff06/powerdnsadmin/templates/dashboard_domain.html#L32) for guidance on how to do this.

milefork:
Hi @zoeller-freinet, ofc, sorry for the mistake.

I've updated the push request.

PR #1078, opened 2021-12-11
fix: making the key name in the config database unique

kate-shine:
The configuration database is basically a key value storage and as such, it will benefit from having unique index on keys.

The configuration keys not being unique also cause issues when populating the settings using automation tools, as the new value gets added instead of replaced, effectively preventing simple upsert operation.

Alternative solution: Dropping the id column altogether and moving PK to name; This can't be easily done in a perfectly database agnostic way though (https://stackoverflow.com/questions/13756567/adding-primary-key-to-existing-mysql-table-in-alembica) and would require a more complex migration. The operation would need a more complex migration script.

This migration might require a manual intervention, if user's database already contains duplicate setting records.


jbe-dw:
Hello,

I personally deploy with Ansible and my role reads the database and chooses the type of operation based on the results. I think your proposal is harmless but a good point. As you open this topic, can someone tell why the settings id's start arround 40+ ?

Finally your patch must be made by Alembic from a change in the model definition, which is not included in the commit.

kate-shine:
Thanks for the note with model definitions, I didn't work with Alembic before :) Fixed. 

As for why does it start with 40+: MySQL has a counter for auto increment tables (you can find it in `information_schema` database, `SELECT AUTO_INCREMENT FROM information_schema.TABLES WHERE TABLE_SCHEMA = "<powerdns_database>" AND TABLE_NAME = "setting";`

That counter doesn't get set to one when all the rows are erased, it just increments. (and when it hits the limit, it just fails and you need to manually reset the autoincrement counter to a lower value to reuse previously erased IDs - `ALTER TABLE setting AUTO_INCREMENT = 1;`)

Your migrations already create some rows there, and delete them all in later migration.

That's one of reasons why it's generally not a good idea to even have an autoincrement ID for a table that already has a unique identifier and that doesn't need an opaque ID (and in that case, UUID is quite often preferred)

It shouldn't be an issue in this case, but if you'd consider a bigger refactoring of those databases at some point, I'll gladly help :)

jdreichmann (approved):

AzorianMatt:
@kate-shine my apologies this took so long but the day has finally come!

PR #1082, opened 2021-12-15
SAML Configuration Interface

vmarkop:
This PR exposes the SAML configuration in the Authentication Settings of the app, accompanied by a detailed Help section, migrating the SAML configuration from the app config to the Settings model in the process.

Minor fixes, such as graceful handling of errors, were also introduced, but the goal of this request is to bring an existing feature to surface and make it more usable, and **NOT** to further improve it.

Developed by [our team](https://github.com/gunet), in collaboration with @kkmanos  and @KostasMparmparousis


jbe-dw (changes requested):
Hello,

I've not strictly checked the whole options but the code is ok for me. I have however two requests regarding this PR:

- Please add the corresponding documentation / or update the existing one to move from settings in config.py to the interface
- Please try to reduce the number of commits in this PR as well

This change in the location of the SAML config is a breaking change and must be considered when merging (and releasing).

jbe-dw:
up

jbe-dw:
up

vmarkop:
Closed / succeeded by PR #1189.

PR #1084, opened 2021-12-16
LDAP Autoprovisioning Update

KostasMparmparousis:
Hello there!
With this Pull Request we aim to overhaul the LDAP Autoprovisioning documentation, to a form that is more readable/understandable. Moreover, we added a more detailed version of the overhaul in the appended project directory for documentations.

In addition to that, we have implemented a minor adjustment to the original provisioning flow, where from now on a Person's Role in the application will be decided by the highest rank (1.Administrator, 2.Operator, 3.User) found amongst his valid Entitlement attributes. Finally, case-insensitive Roles located in the URNs (e.g administrator, USER, etc.) are now also considered as valid.


jbe-dw:
Hello,

I'll give a try. Please consider rebasing once v0.3.0 is released.

Regards

PR #1089, opened 2021-12-25
Enforce Record Restrictions in API

RGanor:
Currently, the administrator can specify which record-types and TTL are allowed within PowerDNS-Admin, but these rules are not enforced anywhere but in the UI rendering.
This PR introduces a check for allowed record-types and TTL in the API.
operators and administrator are exempt from these restrictions.

For future PR, need to add these check to the UI's `apply` API call.


zoeller-freinet:
Glad this caught your attention - thank you for addressing it!

The new decorators currently cause unwarranted Exceptions on `/servers/<string:server_id>/zones/<string:zone_id>` (in cases like this it's handy to have [PowerDNS's OpenAPI spec](https://github.com/PowerDNS/pdns/blob/master/docs/http-api/swagger/authoritative-api-swagger.yaml) at hand to review what the HTTP methods do and how they're meant to be used):
- GET, DELETE
  - `request.get_json()` will return `None`, but the code tries to access `content['rrsets']` later. The new checks do not apply to both methods anyway, so you can skip them on `GET` and `DELETE` (like already implemented by e.g. `apikey_can_remove_domain`).
- PUT
  - The dict returned by `request.get_json()` will not contain the `rrsets` key; the checks are not needed as well (PowerDNS will ignore 'rrsets') and can be skipped, like `GET` and `DELETE`.
- PATCH
  - As long as clients use the API correctly, accessing `content['rrsets']` should work. However, the new decorators still should not raise `KeyError` exception (communicated to the client by `{"msg":"Internal Server Error"}`), if the API is not used correctly. Therefore I'd recommend to wrap both for loops inside `try ... except` statements that catch `TypeError` and `KeyError`, simply `pass`ing inside the except clause. Afterwards it's PowerDNS's job to act correctly if the request body is incomplete or malformed.

TomSebty:
> * However, the new decorators still should not raise `KeyError` exception

Hello, can you please show how you tested this? How did you trigger the `KeyError`?

jbe-dw:
Hello,

The record type enforcement is added in #1113. Still, the TTL enforcement should be an option.

@vmarkop @kkmanos what do you think ?

jbe-dw:
Could you folks give an update here ?

vmarkop:
Hello @jbe-dw!

Indeed, PR #1113, and its successor #1171, fix record-type enforcement, which existed as a thought but was not deeply implemented in the application.

However, after giving the matter some thought, we believe that it is not necessary to offer different TTL options per Role, and that everyone should have access to all available TTL options of a Record, supposing of course that they are allowed to Edit that specific record.

Please let us know if that is not what you meant in your first mention, or if there is an important use case where available TTL options should be further guarded.

Thank you!

jbe-dw:
No the remark was for @RGanor because his intention here is to enforce globally set allowed record and TTLs (the ones in the app setting) in the API.

My comment was that custom roles will enforce the types and that he should change the PR to handle the TTL part. I asked you to confirm that custom roles really enforce allowed record types in the API as proposed here.

Finally, I think that TTL enforcement should be an option to enable in the settings.

vmarkop:
Hello! We can confirm record type permissions are enforced in the API with our PR. We agree with your comment that TTL should be in the settings. I’m glad we see eye to eye, after all!

Thank you.

powerdnsadmin/decorators.py
```diff
@@ -367,6 +367,45 @@ def decorated_function(*args, **kwargs):
...
+    return decorated_function
+
+def allowed_record_ttl(f):
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        if g.apikey.role.name in ['Administrator', 'Operator']:
+            return f(*args, **kwargs)
+
+        allowed_ttls = Setting().get_ttl_options()
+        allowed_numric_ttls = [ ttl[0] for ttl in allowed_ttls ]
```
jbe-dw, line 396:
```suggestion
        allowed_numeric_ttls = [ ttl[0] for ttl in allowed_ttls ]
```

powerdnsadmin/decorators.py
```diff
@@ -367,6 +367,45 @@ def decorated_function(*args, **kwargs):
...
+        content = request.get_json()
+        try:
+            for record in content['rrsets']:
+                if record['ttl'] not in allowed_numric_ttls:
```
jbe-dw, line 400:
```suggestion
                if record['ttl'] not in allowed_numeric_ttls:
```

jbe-dw (changes requested):
Well, there is only one typo. Until custom roles land, I'm ok to add the two checks, however i'm waiting for the toggle setting for the TTL we discussed.

Regards

jbe-dw:
I'd like to add that you must add the request type parsing in the decorator as suggested by @zoeller-freinet because it will be easier for us to understand what really happens in here if these decorators are skipped if it is not a PATCH request.

A quick look at the proposed example will show you that it's quick to implement as well.

PS: don't forget the toggle for the TTL enforcement.
Thank you

PR #1090, opened 2021-12-27
Domain record input safety check

kkmanos:
Hello, 

In this pull request I came up with a
filtering method for the input fields of 
a domain record ("Name", "Data" and "Comment"). 
Preventing user-input abuse, 
will undoubtedly reduce the probability of having any
future vulnerability.

### DIFF
--- a/powerdnsadmin/routes/domain.py
+++ b/powerdnsadmin/routes/domain.py
@@ -666,8 +666,10 @@ def record_apply(domain_name):
         for r in submitted_record:
             if is_forbidden(r['record_comment']):
                 field = "Comment "
-            elif is_forbidden(r['record_data']) and r['record_type'] == 'TXT':
+            elif is_forbidden(r['record_data']):
                 field = "Data field "
+            elif is_forbidden(r['record_name']):
+                field = "Name field "
             else:
                 field = ""
             if field != "":
--- a/powerdnsadmin/routes/domain.py
+++ b/powerdnsadmin/routes/domain.py
@@ -662,7 +662,6 @@ def record_apply(domain_name):
         jdata = request.json
         submitted_serial = jdata['serial']
         submitted_record = jdata['record']
-        submitted_record = jdata['record']
         for r in submitted_record:
             if is_forbidden(r['record_comment']):
                 field = "Comment "
--- a/powerdnsadmin/routes/domain.py
+++ b/powerdnsadmin/routes/domain.py
@@ -645,6 +645,12 @@ def change_account(domain_name):
     else:
         abort(500)
 
+def is_forbidden(s):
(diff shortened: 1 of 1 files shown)


zoeller-freinet:
Hi @kkmanos,

can you describe your threat model / what kind of attack you're trying to defend against with this PR? If it's script injection or cross-site scripting, then particular characters being inserted into particular fields on the back-end side is not an issue.

Rather, problems can emerge on the front-end side if characters that have special meaning inside an HTML context are to be displayed, but are not treated properly for this purpose. There is, for example, no good reason why it shouldn't be possible to use the string `<script>alert(1);</script>` as a comment - it just needs to be displayed as such, instead of executed as javascript! The template engine should make sure that this is the case.

If you think you've spotted a possibility where user-supplied input that is later displayed by a web browser can circumvent the template engine, then the chance would be high that this would be a security vulnerability. In such a case - or any other suspected security vulnerability - please open a respective PR, or better, for responsible disclosure (until I finally get around to arranging an official `security@` email address for PDA), send a mail describing your findings to zoeller [at] freinet.de, I'll try to answer in a timely fashion.

kkmanos:
@zoeller-freinet 
Hello there,
I didn't find any vulnerability that could allow a user to get Administrator or Operator rights. 
But, with the use of a simple window.location.replace() function call, an evil user can redirect a high privileged user
to another site. This site could be a clone of PDNS Admin, for instance.

There are a couple of ways to tackle such a vulnerability, one of them is to convert the modals (in History) into drop-downs, but this changes the whole format of the history log, simply because all the rest of the details of other history log types are displayed in a modal.
I found this to be the most secure way.

Have a happy new year

zoeller-freinet:
Goodness gracious, there were indeed more XSS vulnerabilities in there - thank you for drawing my attention to this. 

The HTML for the modal window was stored inside a value attribute - don't do that :( I didn't review the [history overhaul PR](https://github.com/PowerDNS-Admin/PowerDNS-Admin/pull/1042), and unfortunately didn't notice this was the case, until now. Although the user input was escaped by the template engine, browsers parse character entities as characters [inside attribute values](https://www.w3.org/TR/html401/types.html#type-cdata), and as the value is then read and placed into the DOM via JS, what is output from the server side as e.g. `&lt;script&gt;alert(1);&lt;/script&gt;` is in the end executed as JS.

Prohitibing certain characters from being inserted on the back-end side still wouldn't be a good way to fix this. The core conceptual reason for this is outlined in my previous post. Also, one would consequentially have to prohibit these characters everywhere, not just in the location this PR proposes. Soon the first user would come around asking why they are forbidden from entering certain characters here or there. It would get quite messy quite fast - let's not reinvent an inferior wheel here.

I pushed a hotfix for this (0b2ad52), thanks again and happy new year!

PR #1107, opened 2022-02-02
Set Content-Type for PowerDNS API

larsl-net:
The PowerDNS Admin API requires the content type JSON when changing records, but does not set it even when changing records.

This makes it impossible to create, change or delete records with a PDA that is connected to another PDA via an API key.

### DIFF
--- a/powerdnsadmin/models/record.py
+++ b/powerdnsadmin/models/record.py
@@ -293,7 +293,7 @@ class Record(object):
         return new_rrsets, del_rrsets
 
     def apply_rrsets(self, domain_name, rrsets):
-        headers = {'X-API-Key': self.PDNS_API_KEY}
+        headers = {'X-API-Key': self.PDNS_API_KEY, 'Content-Type': 'application/json'}
         jdata = utils.fetch_json(urljoin(
             self.PDNS_STATS_URL, self.API_EXTENDED_URL +
             '/servers/localhost/zones/{0}'.format(domain_name)),


jbe-dw:
Wow, it's great but i'm wondering what kind of setup needs to chain two PDA ?

larsl-net:
Hi @jbe-dw

I have a PDA (primary) that has direct access to the nodes of my pdns setup, this PDA is connected to my auth system.
A second PDA (secondary) running on a different infrastructure, which uses a different system for authentication.

The connection from the secondary to the primary PDA runs over a user token to limit the users on the secondary to managing the record of selected zones. Thus the secondary PDA has only the possibility to change records at the defined domains (no DNSSEC, no zone creation/deletion).

jbe-dw:
So, if I understood:
- Your main PDA is connected to your PDNS Server and an authentication system.
- Your secondary instance is connected to the main using an API token with a User role on the master, with another authentication system with a limited subset of domains.

This let you have to authentication system of the same kind. As we are far to address this need, I guess it is ok.

There are changes coming in the project management. I will delay your PR a little further to let things be organized before merging.

Regards

jbe-dw:
Hello @larsl-net 

I took the time to test your PR. It turns out that it fix the record creation/update successfully. However, as you explained about chaining PDA instances, this patch looks incomplete. I can't create or update a domain in this scenario, for the same basic reason. Although I havn't tested it, it looks like auto_ptr features is impacted by this bug as well.

I'm gonna open another PR with updated headers soon, after testing its full compatibility with PowerDNS.

jbe-dw:
@larsl-net Can you give a try at #1154 ?

Issue #1011, opened 2021-10-04
AD Login fails if there is a loop in MemberOf recursion.


cropalato:
It is supposed not to have a loop, but if that is the case, the login will fail, because there is no recursion control.

jbe-dw:
Fixed with #1118

PR #1118, opened 2022-02-15
Fixing AD login if there is a infinity loop in memberOf groups.

cropalato:
Fix: AD Login fails if there is a loop in MemberOf recursion. #1011


powerdnsadmin/models/user.py
```diff
@@ -307,7 +307,20 @@ def is_validate(self, method, src_ip='', trust_user=False):
                                                 LDAP_USER_GROUP))
                                     return False
                             elif LDAP_TYPE == 'ad':
-                                user_ldap_groups = []
+                                if LDAP_ADMIN_GROUP:
```
jbe-dw, line 310: Can you initialise all the filters with an empty string and just use a if ?

```suggestion
                                ldap_admin_group_filter, ldap_operator_group, ldap_user_group = "", "", ""
                                if LDAP_ADMIN_GROUP:
                                    ldap_admin_group_filter = "(memberOf:1.2.840.113556.1.4.1941:={0})".format(LDAP_ADMIN_GROUP)
```
Therefore there are many words cut by spaces (maybe your editor) in the PR to fix.

powerdnsadmin/models/user.py
```diff
@@ -317,26 +330,22 @@ def is_validate(self, method, src_ip='', trust_user=False):
                                         .format(self.username))
                                     return False
 
-                                for group in [
-                                        g.decode("utf-8")
-                                        for g in user_ad_member_of
-                                ]:
-                                    user_ldap_groups += self.ad_recursive_groups(
-                                        group)
+                                for i in  range(len(user_ad_member_of)):
```
jbe-dw, line 333: Can you keep the comprehension list ?
```suggestion
                                user_ad_member_of = [g.decode("utf-8") for g in user_ad_member_of]
```

jbe-dw (changes requested):
Hello @cropalato 

Thank you for the PR. The PDA team has changed, we are reviewing the pending PR. I made some comment in your code.

Can you remove `ad_recursive_groups` method as well because it is not used anymore ?

Regards
Jérôme

powerdnsadmin/models/user.py
cropalato: hello @jbe-dw 
Thanks for the feedback. I removed the unused method and replaced the loop syntax. Please let me know if you find something else I have missed.

powerdnsadmin/models/user.py
```diff
@@ -317,26 +303,21 @@ def is_validate(self, method, src_ip='', trust_user=False):
                                         .format(self.username))
                                     return False
 
-                                for group in [
-                                        g.decode("utf-8")
-                                        for g in user_ad_member_of
-                                ]:
-                                    user_ldap_groups += self.ad_recursive_groups(
-                                        group)
+                                user_ad_member_of = [g.decode("utf-8") for q in user_ad_member_of]
```
jbe-dw, line 306: There is a typo here and we are good to go. Can you squash the two changes commit with the next added to fix this line in one and force push so we have two commits to merge ?

I think we'll be good.

cropalato: I believe it should be ok now. Please let me know If you find any other problems.

jbe-dw: I'll check it soon, it looks ok for me thank you.

powerdnsadmin/models/user.py
```diff
@@ -307,7 +285,15 @@ def is_validate(self, method, src_ip='', trust_user=False):
...
                             elif LDAP_TYPE == 'ad':
-                                user_ldap_groups = []
+                                ldap_admin_group_filter, ldap_operator_group, ldap_user_group = "", "", ""
+                                if LDAP_ADMIN_GROUP:
+                                    ldap_admin_group_filter = "(memberOf:1.2.840.113556.1.4.1941:={0})".format(LDAP_ADMIN_GROUP)
+                                if LDAP_OPERATOR_GROUP:
+                                    ldap_operator_group = "(memberOf:1.2.840.113556.1.4.1941:={0})".format(LDAP_OPERATOR_GROUP)
+                                if LDAP_USER_GROUP:
+                                    ldap_user_group = "(memberOf:1.2.840.113556.1.4.1941:={0})".format(LDAP_USER_GROUP)
+                                searchFilter = "(&({0}={1})(|{2}{3}{4}))".format(LDAP_FILTER_USERNAME, self.usern  ame, LDAP_FILTER_GROUP, ldap_admin_group_filter, ldap_operator_group, ldap_user_group)
```
jbe-dw, line 295: Please fix this remaining space issue `self.usern  ame` into the current commit and it will be merged

Issue #1062, opened 2021-12-02
Expired csrf token on login page


jbe-dw:
Hello,

I am often kicked out of PDA after the timeout. It then redirects me on the login page which has a token created at this moment. If I try to log in another timeout later, I have an unauthorized error and have to login again.

Is it a good idea to have an ajax call that request a new token every "timeout - 10s" on the login page or is it a security issue to do so ?

I'm not good on this point, just asking because it happens to me a lot to leave a tab open and have to login twice.

Regards

zoeller-freinet:
I have been running into this issue as well, and have been meaning to work on a fix for this.

jbe-dw:
Or is it even easier to have just a window.reload() loop on the login page ?

PR #1124, opened 2022-02-17
Refresh on csrf token expiration

kkmanos:
Hi, 
This is a fix for the issue #1062


jbe-dw:
Shouldn't it be inside the head tag though ?

bcsanford (commented):
modification works great, no longer get error when i log back in without refreshing page

PR #1127, opened 2022-02-20
Option to forbid the creation of domain if it exists as a record

TomSebty:
This PR presents a new option: `deny_domain_override`. When this option is set to true (ON), users won't be able to create domains if it already exists as a record under an upper domain
For example the user can't create the domain `example.com` if there is a record `example` under `com`

### DIFF
--- a/powerdnsadmin/routes/domain.py
+++ b/powerdnsadmin/routes/domain.py
@@ -31,6 +31,10 @@ domain_bp = Blueprint('domain',
                       template_folder='templates',
                       url_prefix='/domain')
 
+def get_upper_domain(domain_name):
+    upper_domain_name = '.'.join(domain_name.split('.')[1:])
+    print("Upper Domain Name: {}".format(upper_domain_name))
+    return upper_domain_name
 
 @domain_bp.before_request
 def before_request():
@@ -400,6 +404,20 @@ def add():
             account_name = Account().get_name_by_id(account_id)
 
             d = Domain()
+            ### Test if a record same as the domain already exists in an upper level domain
+            deny_domain_override = Setting().get('deny_domain_override')
+            if deny_domain_override:
+                rec = Record()
+                upper_domain_name = get_upper_domain(domain_name)
+                while (upper_domain_name != ''):
+                    if d.get_id_by_name(upper_domain_name) != None:
+                        rrsets = rec.get_rrsets(upper_domain_name)
+                        for r in rrsets:
+                            if r['name'].rstrip('.') == domain_name:
+                                result = {'msg': 'Domain already exists as a record under {}'.format(upper_domain_name)}
+                                return render_template('errors/400.html', msg=result['msg']), 400
+                    upper_domain_name = get_upper_domain(upper_domain_name)
+            ### END CHECK
             result = d.add(domain_name=domain_name,
                            domain_type=domain_type,
                            soa_edit_api=soa_edit_api,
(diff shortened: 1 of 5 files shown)


RGanor:
@zoeller-freinet what do you think about this PR?

TomSebty:
@jbe-dw there is a global setting to toggle so if you don't want to use it you can just disable this feature

The new commit adds the deny to the API and added a setting to override in the domain_add page

vmarkop:
Hello @TomSebty.

Thank you for the PR! We are interesting in merging it. We would further appreciate it if you:

1. Solve the conflicts with the current Master branch, so your PR can be merged.
2. Make a small /docs/DenyDomainOverride.md documentation file, explaining the feature. We are especially interested in you mentioning _why the feature is togglable_ (because of load to the API for domains with big subdomain depth).

Regards,
on behalf of the PowerDNS-Admin team.

powerdnsadmin/routes/domain.py
```diff
@@ -31,6 +31,10 @@
                       template_folder='templates',
                       url_prefix='/domain')
 
+def get_upper_domain(domain_name):
+    upper_domain_name = '.'.join(domain_name.split('.')[1:])
+    print("Upper Domain Name: {}".format(upper_domain_name))
```
RGanor, line 36: Delete this print/ use logger

powerdnsadmin/default_config.py
```diff
@@ -12,10 +12,10 @@
 FILESYSTEM_SESSIONS_ENABLED = False
 
 ### DATABASE CONFIG
-SQLA_DB_USER = 'pda'
```
RGanor, line 15: Delete this unnecessary file changes from your pr

powerdnsadmin/decorators.py
```diff
@@ -278,6 +284,26 @@ def decorated_function(*args, **kwargs):
             msg = "ApiKey #{0} does not have enough privileges to create domain"
             current_app.logger.error(msg.format(g.apikey.id))
             raise NotEnoughPrivileges()
+
+        deny_domain_override = Setting().get('deny_domain_override')
+        req = request.get_json(force=True)
+        override_domain = req["override_domain"]
+        print("Override: {}".format(override_domain))
```
RGanor, line 291: Remove print/use logger

powerdnsadmin/routes/domain.py
```diff
@@ -400,6 +405,19 @@ def add():
             account_name = Account().get_name_by_id(account_id)
 
             d = Domain()
+            ### Test if a record same as the domain already exists in an upper level domain
```
RGanor, line 408: Is there a way to not duplicate this code in the UI and the API?

RGanor (changes requested):

jbe-dw:
@TomSebty Can you push the requested changes ? We asked 30d ago.

RGanor:
Im working on this PR, it is not yet ready to be merged

RGanor (approved):

RGanor:
Updated the checks to also account for the use case of which in domain `a.`, there is a record `b.c.a.` and the user want to create a domain `b.a.`.
Also updated the permissions to only allow admin/op to override records.

powerdnsadmin/routes/domain.py
```diff
@@ -400,6 +399,31 @@ def add():
             account_name = Account().get_name_by_id(account_id)
 
             d = Domain()
+
+            ### Test if a record same as the domain already exists in an upper level domain
+            if Setting().get('deny_domain_override'):
+
+                upper_domain = None
+                domain_override = False
+                domain_override_toggle = True
```
jbe-dw, line 408: Isn't it supposed to be False here ?

jbe-dw (approved):
I want to give a try but the code LGTM

powerdnsadmin/routes/domain.py
RGanor: You are right 🤦

PR #1154, opened 2022-03-30
fix: Set Content-Type on backend API calls

jbe-dw:
Improvment for  #1107

PowerDNS Admin can be configure to interact with another PowerDNS Admin instance's API. In such setup, content can't be managed because PowerDNS makes REST calls without setting a Content-Type header. PowerDNS never complains about this, but PowerDNS Admin needs content type.

This PR adds the Content-Type wherever it may be needed.


larsl-net:
@jbe-dw
I have tested the MR and everything works for me.

PR #1159, opened 2022-04-04
Decode domain record data and comment from HTML entity to text

### LINKED ISSUE #1098: Bug: Special characters in Record data are altered
When Applying Changes to a domain, if an HTML special character (<, & etc.) is present in a Record's data, the special character is replaced by its character reference, even if it was not modified.

This problem is exaggerated by the fact that the "&" character, which is found in all character references, is then recursively replaced with "&amp;" every time changes are applied, which can lead to infinitely long comments.

### Showcase:

Before state:
Domain4

A change to "testrecord" is applied, and record1's comment is corrupted.
Domain6_AfterApplyChanges

"Apply Changes" is clicked once more, which again alters record1's comment
Domain8_AfterApply

After applying another change, the comment keeps growing...
Domain9

AdvanticGmbH:
Resolves #1098

### DIFF
--- a/powerdnsadmin/static/custom/js/custom.js
+++ b/powerdnsadmin/static/custom/js/custom.js
@@ -76,11 +76,15 @@ function getTableData(table) {
         record["record_type"] = r[1].trim();
         record["record_status"] = r[2].trim();
         record["record_ttl"] = r[3].trim();
-        record["record_data"] = r[4].trim();
-        record["record_comment"] = r[5].trim();
+        record["record_data"] = convertHTMLEntityToText(r[4].trim());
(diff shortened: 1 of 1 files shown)


kkmanos:
Hello, thanks for the PR. 
With some testing, I found that PR #1158 solves the issue [#1098](https://github.com/PowerDNS-Admin/PowerDNS-Admin/issues/1098).

Let me know if you find this PR still useful, otherwise I can close this PR.

AdvanticGmbH:
@jbe-dw @kkmanos I can still reproduce the issue with the changes of #1158. If you set the character `<` for domain comment and press apply changes twice the comment will be still altered to `&lt;`.

PR #1163, opened 2022-04-06
Idna decode

### LINKED ISSUE #1162: SOA isnt working with IDNA and domains with "ß" are broken
When the SOA contains a special character like "ü"  apply changes on a Domain throws this:
```
Record test.com./SOA PLACEHOLDER.advantic.net. admin.übung.com. 2022040501 14400 4800 604800 86400: Not in expected format (parsed as PLACEHOLDER.advantic.net. admin.\195\188bung.com. 2022040501 14400 4800 604800 86400)
```

Also does any domain containing the character "ß" crash the dashboard.
```
<button type="button" class="btn btn-flat btn-success button_template" id="{{ domain.name }}">
File "PowerDNS-Admin/powerdnsadmin/lib/utils.py", line 253, in pretty_domain_name
raise Exception("Cannot decode IDN domain")
```

AdvanticGmbH:
Resolves #1162

### DIFF
--- a/powerdnsadmin/lib/utils.py
+++ b/powerdnsadmin/lib/utils.py
@@ -4,6 +4,7 @@ import json
 import requests
 import hashlib
 import ipaddress
+import idna
 
 from collections.abc import Iterable
 from distutils.version import StrictVersion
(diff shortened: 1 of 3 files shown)


jbe-dw (changes requested):
Please don't use double quotes. I know that it is messy in this project, but single quotes are used arround the changes.

Another thing is idna is not present in Python 2 by default. Nothing says that we dropped Python 2 support yet. @RGanor @vmarkop @kkmanos should we ask idna in requirements.txt ?

Thanks for the PR

RGanor:
@jbe-dw , im not an expert in idna, but I dont understand why we need this dependency and how doest it differ from "string".encode("idna"). @AdvanticGmbH can you please elaborate on this topic?

AdvanticGmbH:
> @jbe-dw , im not an expert in idna, but I dont understand why we need this dependency and how doest it differ from "string".encode("idna"). @AdvanticGmbH can you please elaborate on this topic?

The problem with this is that the library behind `encode("idna")` doesn't work with the German character "ß". If any Domain name contains this character the library throws an error, although "ß" is an valid idna character. For testing this, add an domain with "ß" in the name. As soon the dashboard is shown an error will be shown because `encode("idna")` is currently used to pretty print the names for the dashboard. The error might look like this:
`DataTables warning: table id=tbl_domain_list_reverse - Ajax error. For more information about this error, please see http://datatables.net/tn/7`

AdvanticGmbH:
@jbe-dw I have added your improvements. Do we want to add `idna<3` to the requirements.txt since version 2.x is the supported version for Python 2?

jbe-dw (approved):

jbe-dw:
No, we've talked about this and we think that PDA is probably broken in Python2 now.

AdvanticGmbH:
@jbe-dw I think this is ready to get merged, if there aren't any showstoppers anymore.

jbe-dw:
As your other patch caused some trouble because older history entries are not compatible with a raw json.loads, I prefer to focus on a good patch.

I need some time to double check the issue and the patch. I'm not using IDNA either so it is not straightforward to me. But it should land soon

PR #1182, opened 2022-04-27
fix: Insert valid JSON in history.detail and replace single quotes in the database

jbe-dw:
#1160 removed some character replacement that are needed for databases with single quotes in the history table. 

Users with such entries in their history are not able to use PDA anymore. This small patch is for testing until a more robust solution is pushed.

Regards

### DIFF
--- a/migrations/versions/fbc7cf864b24_update_history_detail_quotes.py
+++ b/migrations/versions/fbc7cf864b24_update_history_detail_quotes.py
@@ -1,0 +1,47 @@
+"""update history detail quotes
+
+Revision ID: fbc7cf864b24
+Revises: 0967658d9c0d
+Create Date: 2022-05-04 19:49:54.054285
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = 'fbc7cf864b24'
+down_revision = '0967658d9c0d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    history_table = sa.sql.table(
+        'history',
+        sa.Column('id', sa.Integer),
+        sa.Column('msg', sa.String),
+        sa.Column('detail', sa.Text),
+        sa.Column('created_by', sa.String),
+        sa.Column('created_on', sa.DateTime),
+        sa.Column('domain_id', sa.Integer)
+    )
+
+    op.execute(
+        history_table.update().where(
+            sa.and_(
+                history_table.c.detail.like("%'%"),
+                history_table.c.detail.notlike("%rrests%"),
+                history_table.c.detail.notlike("%rrsets%")
+            )
+        ).values({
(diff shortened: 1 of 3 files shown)


powerdnsadmin/routes/admin.py
```diff
@@ -93,7 +93,7 @@ def extract_changelogs_from_a_history_entry(out_changes, history_entry, change_n
         return
 
     if "add_rrests" in history_entry.detail:
-        detail_dict = json.loads(history_entry.detail)
+        detail_dict = json.loads(history_entry.detail.replace("\'", ''))
```
jbe-dw, line 96: @vmarkop @kkmanos You worked on detail history, can you remind us why we added this replacement ?

Thanks

vmarkop: Hello,

the above command removes any single quotes that were escaped with a backslash, as their existence used to break history rendering and throw errors.

Thank you

powerdnsadmin/routes/admin.py
```diff
@@ -763,7 +763,7 @@ def __init__(self, history, change_set):
             self.detailed_msg = ""
             return
 
-        detail_dict = json.loads(history.detail)
+        detail_dict = json.loads(history.detail.replace("'", '"'))
```
jbe-dw, line 766: I believe this one should not be reverted, but we should push an alembic script to replace quotes in all history entries instead.

powerdnsadmin/routes/admin.py
AdvanticGmbH: @jbe-dw I cant reproduce that issue on my local environment. How can I cause it?

jbe-dw: Hello,

Because before your patch history details are written with single quotes. json.loads crashes because it is not a valid JSON.

You can try to update one of the 4 most recent history entry (like successfull login) in your db then reload the dashboard.

jbe-dw: @vmarkop 

I have worked on the Alembic script pushed in this PR. It should solves the corrupted quote issues fixed L766.

For line 96, I need some corrupted history entries addressed by this interpolation to produce a clean patch as well.

Considering that it was added by your team, I really rely on you to gather it (either from one of your managed instance or maybe a ticket open by your customer).

I think it is far better to fix the db log rather than patching the python code. As @AdvanticGmbH, we should store valid JSON as we can in text columns.

If you can take some time to test the migration script on your side it would be appreciated.

jbe-dw:
Hello,

The comment are outdated now, if you look at the patch we opted for trying to store valid JSON in the detail history column. I think I have changed the missing calls in the current version, including the one you pointed.

There is a db migration script to apply too, to replace single quotes with double quotes. We don't have corrupted history entries that are not fixed with this.

PR #1188, opened 2022-05-07
Small bugfixes

### LINKED ISSUE #1086: Drop pyOpenSSL dependency
https://pypi.org/project/pyOpenSSL/ states:

> Note: The Python Cryptographic Authority strongly suggests the use of pyca/cryptography where possible. If you are using pyOpenSSL for anything other than making a TLS connection you should move to cryptography and drop your pyOpenSSL dependency.

pyOpenSSL depends on [pyca/cryptography](https://pypi.org/project/cryptography/) itself. In PowerDNS-Admin, pyOpenSSL is only used inside [`lib.certutil`](https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/7808febad8fb0a56cdecce2d7cdb9a92ec30ff06/powerdnsadmin/lib/certutil.py); `check_certificate` is never called and `create_self_signed_cert` does nothing that couldn't be done with `cryptography`.

corubba:
This PR is a collection of a few small bugfixes. Below is a full digest of the commit messages.

## Fix revision in migration filename
This has no functional impact, flask-migrate aka alembic was and will continue to work as expected. It is just a cosmetic change for consistency.

## Remove python v2 remnant
As [vermin](https://pypi.org/project/vermin/) confirms, the codebase has long moved beyond supporting python v2 (which is not a bad thing). This removes the last explicit py2 piece of code.

And in case anyone wonders, vermin currently reports the minimum version to be v3.6.

## Header for fixed order column
Semantically and syntactically it is better to have the same number of `<th>` as `<td>`. Not that anyone will ever see that new header, since that column is always invisible (except if the user disables javascript).

Plus remove a unmatched closing html element.

## Replace pyOpenSSL with cryptography
This is literally the example from the [docs](https://cryptography.io/en/latest/x509/tutorial/#creating-a-self-signed-certificate). The only thing I adapted are the parameters for the keys and certificate, so they stay the same.

Fixes #1086


jbe-dw:
Hello,

Thanks for your contribution, we will validate it soon, but the code is ok for me.

Regards

jbe-dw:
Hello,

Thanks for your contribution, we will validate it soon, but the code is ok for me.

Regards

jbe-dw:
Hello,

Since you're doing interesting PR, and that your code looks pretty clean to me, I'm asking you if you would agree to help us to improve the project (there's a lot of things pending) by either contributing to pending issues or joining the team. Your profile is quite private and I have no address to contact you outside GitHub.

Let me know if you are interested. You can join us by email at powerdnsadm [at] gmail [dot] com if you don't want to disclose your address here.

Regards

jbe-dw (approved):
LGTM

jbe-dw (approved):

jbe-dw (approved):
Ok

PR #1189, opened 2022-05-10
SAML Configuration User Interface

vmarkop:
Hello,

this PR is an update to PR #1082, to fetch the latest changes in Master, clear the code and provide documentation.

As stated in the previous request, this PR exposes the SAML configuration in the Authentication Settings of the app, accompanied by a detailed Help section, while migrating the SAML configuration from the app config to the Settings model in the process.

Minor fixes, such as graceful handling of errors, were also introduced, but the goal of this request is to bring an existing feature to surface and make it more usable, and NOT to further improve it.

Feedback, including bug-testing and constructive critisism on the documentation, is accepted and kindly requested.

Developed by [our team](https://github.com/gunet), in collaboration with @kkmanos  and @KostasMparmparousis


jbe-dw:
Thanks for the update.

We should ask SAML users to give a test because I don't have a SAML setup yet.

I'll take another look within the week.

pixelrebel:
This is a great feature!  I tested this and noted a few minor issues:

1.  SAML errors (i.e. invalid_response) no longer print on the SAML.html template. 
2. It could have been my browser cache, but after several updates to the config, the fields stop updating.  I would change a value in several fields, after saving, the values in the fields reverted back to their original values.  
3. Some of the non-required fields populate with the example data instead of leaving it blank.  
4. The example config in `development.py` is incomplete and thus some of the variables like `saml_attribute_group` and `saml_group_admin_name` haven't been ported to the UI yet.  I have also exposed a couple more settings  with instructions in [PR #1203](https://github.com/PowerDNS-Admin/PowerDNS-Admin/pull/1203).

Excellent work!

vmarkop:
Hello @pixelrebel,

Thank you for your feedback! This PR will be updated as soon as SAML-related pull requests #1203 and #1205 are merged, to fix the issues you have mentioned, as well as implement the changes and documentation of the above pull requests in the UI.

> This is a great feature! I tested this and noted a few minor issues:
> 
> 1. SAML errors (i.e. invalid_response) no longer print on the SAML.html template.
> 2. It could have been my browser cache, but after several updates to the config, the fields stop updating.  I would change a value in several fields, after saving, the values in the fields reverted back to their original values.
> 3. Some of the non-required fields populate with the example data instead of leaving it blank.
> 4. The example config in `development.py` is incomplete and thus some of the variables like `saml_attribute_group` and `saml_group_admin_name` haven't been ported to the UI yet.  I have also exposed a couple more settings  with instructions in [PR #1203](https://github.com/PowerDNS-Admin/PowerDNS-Admin/pull/1203).
> 
> Excellent work!

powerdnsadmin/routes/index.py
```diff
@@ -592,26 +592,34 @@ def prepare_welcome_user(user_id):
...
+                current_app.logger.info(
+                "SAML: Your IDP does not support Single Logout.")
+        try:
             return redirect(
                 auth.logout(
                     name_id_format=
                     "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress",
                     return_to=current_app.config.get('SAML_LOGOUT_URL'),
                     session_index=session['samlSessionIndex'],
                     name_id=session['samlNameId']))
```
jbe-dw, line 619: I don't understand why this stays. You check if there is a logout url in the settings above. If not, it means it should not be tried, doesn't it ?

I also note that you're calling current_app.config.get('SAML_LOGOUT_URL'), it should be fetched from settings as well.

jbe-dw (dismissed):
Well, I've just made a comment on some lines but everything else looks ok to me. TBH as I don't have an IDP to test it, I can't give feedback on the code behaving. But looking at it didn't raise any doubt it would work.

powerdnsadmin/routes/index.py
vmarkop: Hello,

this old block of code slipped through us while rewriting the PR, that's why the current_app.config.get() statement is still there.

Thank you for pointing this out!

hapylestat:
@AzorianMatt just to know - upcoming in which year?

AzorianMatt:
> @AzorianMatt just to know - upcoming in which year? 

@hapylestat Presumably 2025. 🙂

PR #1195, opened 2022-05-17
Adding venv and yarn-error.log to gitignore

Metrax:
venv: in the wiki, the installation is described with creating the venv into "venv", but only ".venv" is in gitignore
yarn-error.log: file is created if yarn fails, it should not be commited to the repo accidentally

This is a preparation for some more pull requests of me soon.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,5 +38,7 @@ node_modules
 powerdnsadmin/static/generated
 .webassets-cache
 .venv*
+venv*
 .pytest_cache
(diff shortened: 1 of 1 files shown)


jbe-dw:
Thanks for contributing.

If PR seems stuck, it is because we lack some time to better test them before merging. The simple ones could be added quickly but larger ones will not be merged that fast.

Anyhow, all help to make PDA become better is really appreciated.

PR #1198, opened 2022-05-18
Fix tests

corubba:
The tests were in a sad state, and still are. But at least they run and pass again.

## Fix test docker
The Dockerfile did not work as is, because the dependencies in `requirements.txt` are newer than the stretch-image with its python v3.5 can support/run. Use stable debian with the lts nodejs instead, plus had to add some libs to make the wheel build succeed.
jsonschema v4 breaks things, so its version needs to be pinned until [bravado is fixed](https://github.com/Yelp/bravado-core/pull/385/files#r731674447).

## Fix tests
Increased the version of pytest to [make it work with py 3.10](https://github.com/pytest-dev/pytest/pull/8540). 
The GET calls no longer return list but the object itself, fixed the tests and assertions to account for that. The tests did not account for the later added `allow_user_remove_domain` setting. And there were issues with missing and non-stopped patchers/mocks.


jbe-dw:
Hello,

To be transparent we are not great at test now. It is something that has to be improved all the way though. We keep it aside now (don't worry if not merged quickly).

I've set up milestones about some opened PR that may need to be included in the test sequence as well.

Please let us know your feelings about this, if there's not much extra work for you to check and add them. I understand if it's not in your priorities/perimeter.

Regards

jbe-dw:
I agree with you, but the project is not yet prepared for such things. Tests should be here to ensure that nothing breaks. But as you could probably see there is no rule about formatting or testing, or even cleanly providing good git PR.

We will probably focus on this aspect at some point. I'll try the test soon and propose a corrective PR afterwards.

Thanks.

AzorianMatt:
This PR is superseded by PR [1339](https://github.com/PowerDNS-Admin/PowerDNS-Admin/pull/1339) so I'm closing it.

PR #1199, opened 2022-05-18
Fix rrest typo in history detail

corubba:
There is a misspelling of rrset throughout the history logic, which also effects the json payload in the `history` database table. The migration will fix the existing payloads, so there is no need for backwards-compatibility and the code-change is a simple search-and-replace.

I am not really happy with the migration, but it is the best I could come up with to modify the json payload in a database-agnostic manner. Any suggestion are welcome.


vmarkop:
Hello @corubba,

your migration script is great, and tests have proven that it works even with old databases.

However, I noticed that on powerdnsadmin/models/record.py lines 362 and 365, the variable `rollback_rrests` still exists.

Could you please fix that as well, so the typo is erased from everywhere? The feature can be merged when this is done.

Thank you!

PR #1201, opened 2022-05-18
Update zone using a single api call

corubba:
Starting with the very first commit, the update was always done with two api calls: one for DELETE and one for REPLACE. It is however perfectly valid and save to do both at once, which makes it atomic, so no need for the rollback. Plus it only updates the serial of the zone once.
There is no point in sending the full RRset data when deleting it, the key attributes to identify it are enough. This also makes the behaviour consistent with the [api docs](https://doc.powerdns.com/authoritative/http-api/zone.html#rrset) where it says "MUST NOT be included when changetype is set to DELETE."

### DIFF
--- a/powerdnsadmin/models/record.py
+++ b/powerdnsadmin/models/record.py
@@ -303,10 +303,47 @@ class Record(object):
                                   data=rrsets)
         return jdata
 
+    @staticmethod
+    def to_api_payload(new_rrsets, del_rrsets):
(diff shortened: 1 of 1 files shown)


jbe-dw:
Hello,

Thanks for your commit. I have successfully tested it. I think it would have been better to reword `Record.compare()`, but we may improve it in a future version.

PR #1203, opened 2022-05-20
Small fixes to SAML service

pixelrebel:
This PR addresses four issues with the SAML service:

1. The example `development.py` config had incorrect labels for `SAML_CERT` and `SAML_KEY`
2. For users getting started with a simple proof-of-concept, not having to configure AttributeStatements on the IdP side would be helpful.  To make it easier, `SAML_WANT_ATTRIBUTE_STATEMENT` should probably default to `False`, but at least having it exposed in the config with an explanation will go a long way for beginners.
3. The current SAML service does not work when run with gunicorn behind an SSL proxy web server.  All of the relay, and response URLs end up hard-coded with `http://`.  Taking a hint from the comment in the `prepare_flask_request` method, we use the proxy HTTP headers to tell us if we should build the URLs with `http://` or `'https://`.  (I also updated the nginx example in the wiki to include the `X-Forwarded-Proto` header so this will work out of the box for users.  
4. Using `SAML_ATTRIBUTE_GROUP` in combination with `SAML_GROUP_ADMIN_NAME` is an excellent method for elevating users to admin by group membership from the SAML assertion.  Added documentation in the `development.py` config to help users get started with this approach.

### DIFF
--- a/configs/development.py
+++ b/configs/development.py
@@ -148,6 +148,10 @@ SAML_ENABLED = False
 
 # #SAML_ASSERTION_ENCRYPTED = True
 
+# Some IdPs, like Okta, do not return Attribute Statements by default
+# Set the following to False if you are using Okta and not manually configuring Attribute Statements
+# #SAML_WANT_ATTRIBUTE_STATEMENT = True
+
 # Remote authentication settings
 
 # Whether to enable remote user authentication or not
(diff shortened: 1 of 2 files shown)


vmarkop:
Hello, @pixelrebel!

Thank you for your work. Your fixes will be merged into the codebase, and pending PR #1189 will be updated to move your new documentation in the Help sidebar and the option `SAML_WANT_ATTRIBUTE_STATEMENT` in the UI Settings. We hope you are okay with this.

Thank you

PR #1211, opened 2022-05-27
Fix cookie config

### LINKED ISSUE #906: The cookie are missing the "secure" attribute
In our vulnerability checks we found the transfer of a session cookie via http on the / route. The following is a complete report to assist in the correction. **We recommend the possibility to enable the secure and same site tag.**

### **Report** 

Vulnerability Detection Result:

The cookies:

Set-Cookie: _csrf_token=1b95889f5f4b9d930b8d80dc8a482c5b88c53819; Expires=Wed, 03-Mar-2021 14:53:12 GMT; Max-Age=***replaced***; Path=/
Set-Cookie: session=.eJwNyF0KwjAQReG93GfB1jZwJ5sJ-ZlBaI2SRBDEvZu3850vQu7NwngeWuGxJnGkmLM9SZFtSSxcSmbcecsukdltXAUXBGva7_AWz66TL22PWLUO-NHe81T9zMYVvz9NVB-Y.YDkLWA.0VBhRJjlmUG9DPjoaE-E7btIcNk; Expires=Fri, 26-Feb-2021 15:03:12 GMT; HttpOnly; Path=/

are missing the "secure" attribute.

Solution:

Set the 'secure' attribute for any cookies that are sent over a SSL/TLS connection.

Affected Software/OS:

Server with SSL/TLS.

Summary:

The host is running a server with SSL/TLS and is prone to information
disclosure vulnerability.

Insight:

The flaw is due to cookie is not using 'secure' attribute, which
allows cookie to be passed to the server by the client over non-secure channels (http) and allows attacker
to conduct session hijacking attacks.

Impact Level: Application

CVSS Base Vector:

AV:N/AC:L/Au:N/C:P/I:P/A:N

References:

https://www.owasp.org/index.php/SecureFlag
http://www.ietf.org/rfc/rfc2965.txt
https://www.owasp.org/index.php/Testing_for_cookies_attributes_(OWASP-SM-002)

CVSS Base Score: 6.4
SID:1

### LINKED ISSUE #907: The cookies are missing the "httpOnly" attribute.
In our vulnerability checks we found the transfer of a session cookie via http on the / route. The following is a complete report to assist in the correction. We recommend the possibility to enable the secure and same site tag.

### Report:
Vulnerability Detection Result:

The cookies:

Set-Cookie: _csrf_token=1b95889f5f4b9d930b8d80dc8a482c5b88c53819; Expires=Wed, 03-Mar-2021 14:53:12 GMT; Max-Age=***replaced***; Path=/

are missing the "httpOnly" attribute.

Solution:

Set the 'httpOnly' attribute for any session cookie.

CVSS Base Vector:

AV:N/AC:L/Au:N/C:P/I:N/A:N

Insight:

The flaw is due to a cookie is not using the 'httpOnly' attribute. This
allows a cookie to be accessed by JavaScript which could lead to session hijacking attacks.

Summary:

The application is missing the 'httpOnly' cookie attribute

Vulnerability Detection Method:

Check all cookies sent by the application for a missing 'httpOnly' attribute

Affected Software/OS:

Application with session handling in cookies.

References:

https://www.owasp.org/index.php/HttpOnly
https://www.owasp.org/index.php/Testing_for_cookies_attributes_(OTG-SESS-002)

CVSS Base Score: 5.0
SID:1

corubba:
This PR contains a few bugfixes related to the cookie configuration. Below is a digest of the full commit messages.

## Fix csrf configuration
The CSRF component has been initialized *before* the app config was fully read, which made it impossible to configure CSRF properly. Moved the CSRF initialisation into the `routes` module where it better fits.
Since I was at it I switched from programmatic to decorated exemptions. GET routes don't need to be exempted because they are by default.

## Set csrf cookie to httponly
The CSRF token is currently inserted directly in the template and not in the browser via JavaScript from the cookie, so making it inaccessible is not a problem.

The Sesson-cookie has `httponly` already set [by default](https://flask.palletsprojects.com/en/2.1.x/config/?highlight=session_cookie_httponly#SESSION_COOKIE_HTTPONLY).

Fixes #907 

## Allow secure cookie config in docker
Setting the two options `SESSION_COOKIE_SECURE` and `CSRF_COOKIE_SECURE` to `True` is recommended if (and only if) you serve PDA via HTTPS. It will break things on plain-HTTP deployments. For non-Docker deployments these can be set in the flask config file, for docker they have to be whitelisted to be set via env vars.

Fixes #906


powerdnsadmin/default_config.py
```diff
@@ -10,6 +10,7 @@
 HSTS_ENABLED = False
 OFFLINE_MODE = False
 FILESYSTEM_SESSIONS_ENABLED = False
+CSRF_COOKIE_HTTPONLY = True
```
jbe-dw, line 13: The only thing I can see is that people will already have a config file missing this setting. Is it possible to enforce it with a default to 'True' when the app init ?

jbe-dw (changes requested):
Thanks for the quality of the submission, I have made one comment.

Nice work

jbe-dw (approved):

Issue #174, opened 2017-01-08
Remove includes of external sites


mathieumd:
PowerDNS-Admin is bound to be used inside enterprises premises, managing very sensible data. It should not include any links to providers like `fonts.googleapis.com`, `cdnjs.cloudflare.com` or `gravatar.com`! All such data should be included into the sources, or if not possible (for Gravatar) enabled only _after_ confirmation by the user.

lethargosapatheia:
DNS is part of core infrastructure. I can't believe this is the default, even after 4 years since the issue was opened! We have the exact same problem, we deal with real sensitive (not sensible) data. That we would give google hints about when we access our dns interface is outrageous, I'm sorry to say. This is a big confidentiality/trustworthiness problem.

danimo:
`OFFLINE_MODE = True` removes requests to Google and Gravatar with no other disadvantages.

That being said, I dislike the wording (and the default), because it implies that contacting 3rd party services for fonts and pictures is at all desirable. HTTP/2 obsoletes the performance advantage, and traffic should be negligible for a DNS frontend. Also, images can be delivered as claims in a more data protection friendly manner when using e.g. OIDC.

My recommendation would be to:

1. Kill the code path that uses Google for fonts entirely
2. Support OIDC claims for pictures
3. Make Gravatar an optionally supported feature
4. Remove OFFLINE_MODE

If  @ngoduykhanh agrees, I would draft a PR for this. If not, I would draft one explaining offline mode in `README.md`.

zoeller-freinet:
Hi @danimo,

given #1032, I'm inclined to say: please go ahead. With @RGanor and I, there are at least two maintainers now, so we should be able to make it happen.

RGanor:
@danimo what you say make sense, but bare in mind that offline mode is not just for the reasons you mentioned above, some deployments are completely isolated from the internet, so relying on external service is not possible.
If any changes will be made, it still need to support such deployments.

danimo:
I am aware of the offline modes intent, but with a privacy-by-default mode, the whole point is that there should no outside services left to contact in a default deployment.

jbe-dw:
Hello @danimo 

I agree with you. We don't have resources to adress it right away but it should be done in a not too far future.

Regards

PR #1212, opened 2022-05-27
Privacy first

corubba:
One of the first surprises for me was that PDA has a offline mode, which is not active by default. Inspired by #174, this PR aims to move PDA in the direction of privacy-by-default.

It is intentional a draft, as there are a few open question.

## Always use local fonts
They are served as static assets, so they are always available and also properly cached in the browser.
I don't like that the associated files are all commited to git repo, and tried to pull them in via yarn instead; but unfortunately couldn't get it to work.

## IE8 polyfills
There are conditional resources to support IE8 and older, which are always fetched from the internet. These old browsers are [EOL since 2016](https://web.archive.org/web/20160115070611/https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support
). There are currently two commits for this: The first allows PDA to serve the polyfills itself, whereas the second simply removes the polyfills and thus IE8 support altogether.

Question to the maintainers: Do you want to keep supporting these old browsers? If yes then I will remove the second commit, if no then I squash them together.

## Use the doc search directly
In the spirit of privacy, the documentation links for PDNS statistics and config options are changed from a google search to the search of the documentation itself.

To be honest, I am not really happy with the result quality of the documentation-included search. Unfortunately there is no protocol for searching with the default search engine of the user, like with `mailto` for sending emails. Maybe using a privacy-oriented search engine like [DuckDuckGo](https://duckduckgo.com/) would be the better solution.

Question to the maintainers: Any opinions which solution you would prefere?

## Rework user image handling
Moved all the logic out of the template into a separate endpoint. This makes it easy to extend to also support images from different sources like LDAP/SAML/OIDC. Session-based caching is hard to do, so to allow time-based caching in the browser, the url needs to be unique for every user by using a query parameter.

Replaced the default/fallback user image with a new one. It is based on the old one, but does not need css to be visible. And removed said css.

Gravatar has now its own setting named `gravatar_enabled`, which is disabled by default.

## Remove `OFFLINE_MODE` config option
At this point this option does nothing anymore, so remove it.


mathieumd:
Thanks @corubba to address this problem!

I don't think supporting very old browsers is worth the effort, but I may be wrong of course.

It's nice to use PowerDNS' doc internal search. It may be broken in the future if they change their site, but it's still better than using Google or any other third party, IMHO.

AdvanticGmbH:
@AzorianMatt What is the status of this pull request? I think that the privacy topic is really important. Can we merge this pull request in the near future?

AzorianMatt:
@AdvanticGmbH I'll move this up on my priority list to get addresses by next month.

PR #1213, opened 2022-05-28
Update Javascript libraries

### LINKED ISSUE #1074: jQuery Missing From NPM Package / Flask Asset Build Process Yields Broken JS
I'm currently building from the master branch which has two issues that prevent a clean build for local testing.

1) The NPM package file (/package.json) is missing a reference to the jQuery NPM package. This results in a failure of the Flask asset building process since the source file is not present locally. I rectified this issue by adding the "jquery" package locked at version 3.6.0 which of course will have to be tested and possibly adjusted.

2) The Flask asset build process is yielding a mistake that causes JavaScript errors in the console, resulting in a broken dashboard page. At line 4661 in my generated main.js file, the line breaks prematurely, leaving the remainder of the line's contents on line 4662. I'm not sure of the cause of this but it is a result of the jsmin concatenating filter being applied. I did notice it's from the icheck NPM package. Unfortunately, this package has only had one release version ever and that was seven years ago. If this is an error in the upstream repository, then this functionality needs stripped and/replaced. I'm quite certain whatever the desired results, it can be achieved using the already in place UI libraries.

jbe-dw:
This PR includes all 4 PR opened by dependabot. We know that we need to have a lot more work pending to update all libraries, but since it is not our priority, we can include these ones because they provide security fix.

Python `jsmin` is outdated and can't minify some regexp correctly. I switched to `rjsmin` which is ... say more recently updated. Note that it allows to use `npm install` and not yarn (which would not be required anymore) at a first glance. It has to be more tested though.

Using `npm install` with the provided `package.json`  and `jsmin` doesn't work because it resolves dependencies differently from yarn (i think that jquery fetched version is different) and encounter jsmin bug.

Fixes #1074
Includes #986 #1031 #1103 #1164

### DIFF
--- a/powerdnsadmin/assets.py
+++ b/powerdnsadmin/assets.py
@@ -24,7 +24,7 @@ js_login = Bundle('node_modules/jquery/dist/jquery.js',
                   'node_modules/bootstrap/dist/js/bootstrap.js',
                   'node_modules/icheck/icheck.js',
                   'custom/js/custom.js',
-                  filters=(ConcatFilter, 'jsmin'),
+                  filters=(ConcatFilter, 'rjsmin'),
                   output='generated/login.js')
 
 js_validation = Bundle('node_modules/bootstrap-validator/dist/validator.js',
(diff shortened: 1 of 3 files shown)


jbe-dw:
@azoriansolutions if you have a thought it is welcome.

AzorianSolutions:
> @azoriansolutions if you have a thought it is welcome.

Seems like a good idea to me. Last time I had my hands in the project, seems like I remember having this problem and quickly finding that it was not going to be an easy solve. Seems like you've got a reasonably quick solution to get past it.

jbe-dw:
@RGanor @vmarkop Can you give a try, so it could be merge in the next release ?

PR #1214, opened 2022-05-30
Fix DynDNS2 using X-Forwarded-For

gadall:
When a DynDNS request is submitted with an empty myip argument, and an X-Forwarded-For HTTP header is present, nothing is done. This fixes the issue.

### DIFF
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -836,7 +836,7 @@ def dyndns_update():
 
     remote_addr = utils.validate_ipaddress(
         request.headers.get('X-Forwarded-For',
-                            request.remote_addr).split(', ')[:1])
+                            request.remote_addr).split(', ')[0])
 
     response = 'nochg'
     for ip in myip_addr or remote_addr:


RGanor:
I didnt really work with dyndns in pdnsadmin. Can you give an practical example on how to replicate this issue?

gadall:
@RGanor 

- Set up pda behind nginx as documented, X-Forwarded-For being set.
- Call /nic/update?hostname=some.host (or /nic/update?hostname=some.host&myip=)

gadall:
If no X-Forwarded-For header will be added upstream you can just add one e.g. curl -H 'X-Forwarded-For: 1.2.3.4'

RGanor:
I dont think i understand the issue yet.
If there is xff header, return it, if not return the remote address. Why this split is even necessary ?
And it will be very helpful if you'll post here an example request (URL + body)

gadall:
You can't just return it. It contains one or more IP address, separated by comma and, in fact,  _optional_ surrounding spaces. You're also supposed to test whether each IP address is private and use the first none-private one.
If exists, each proxy prepends the client's remote address to the existing X-Forwarded-For. This is why the most simplistic approach is split(', ')[0] but a regex with '\s*,\s*' (PCRE) would be more conformant.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

The issue is extremely simple. I already posted a sample request. What do you mean by body? Response body?

gadall:
See other usages of X-Forwarded-For in this same application.

gadall:
> prepends

appends. sorry.

RGanor:
LGTM, merged.

PR #1218, opened 2022-06-13
Improve performance of domain update

AdvanticGmbH:
* Previously the loop would always query the db even for none existing accounts
* Now cache previously looked up accounts and skip db queries for empty accounts

Loop time previously: 3.000699043273926 sec
Loop time now: 0.2798025608062744 sec

### DIFF
--- a/powerdnsadmin/models/domain.py
+++ b/powerdnsadmin/models/domain.py
@@ -143,9 +143,20 @@ class Domain(db.Model):
                 current_app.logger.debug(traceback.format_exc())
 
             # update/add new domain
+            account_cache = {}
             for data in jdata:
                 if 'account' in data:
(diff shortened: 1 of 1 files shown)


AdvanticGmbH:
This drastically increases the loading time of the dashboard for us with around 1000 dns zones.
Maybe you could take a look at it @jbe-dw

jbe-dw:
Nice one. You meant "decreases" I guess :p

I'm sorry we are not as fast as we intended for merging, but we should move on this month.

AdvanticGmbH:
Yes, that is what I meant. ^^

RGanor:
@AdvanticGmbH The numbers are very impressive! I look at this soon.
Thank you for the contribution!

PR #1227, opened 2022-06-23
Allow new domains to be absolute

corubba:
Allow the new domain name to be input absolute (with a dot at the end). To keep the rest of the logic working as-is, remove it fairly early in the function.

Would have loved to use `str.removesuffix()` but that's python v3.9+ only.


jbe-dw:
Hi,

Thanks for this PR. In my local code I use rstrip('.') (but this is fine as well)

PR #1231, opened 2022-06-28
Allow to manage associated domains under account edit

AdvanticGmbH:


AdvanticGmbH:
Also does this patches fix the case that if you would call the route "/admin/account/edit/none_existing" with an none existing account name an 500 error would occur. Now it directs to an fresh empty edit page.

AdvanticGmbH:
@jbe-dw I don't know if you already want to take a look at this since this is more like a feature.

jbe-dw:
Hello,

Thank you for the PR. I let @RGanor @vmarkop take time to review it because I'll be on holiday tomorrow. As our policy said it's not going to be in the priorities, but it's a small one.

We should think if it is relevant to allow the same behaviour on the API. It's unlinked now because this association is made when editing domain on the PowerDNS API (not PDA's).

@corubba what's your opinion about the API ?

Regards

AdvanticGmbH:
@AzorianMatt Is there a chance that this can be added as a feature to the the next batch of merges?

AzorianMatt:
@AdvanticGmbH I should be able to manage that. It's my hope to start merging some of these throughout the next week once I return from vacation.

PR #1238, opened 2022-07-27
Fix LDAP user group search for nested groups

melck:
Hello,

We use an AD with nested users group for groups ADMIN, OPERATOR and USER.

Current code will find a user with `searchFilter` on the 3 groups but it will failed when looking on `memberOf` attribute, because it do not contains nested groups.

The current fix will work in this case but will do 3 requests max.

We could filter nested groups of the user but we would need to changes settings to introduces a base dn on groups.

I think its better for moment to make more requests to ldap than introduces breaking changes

What do you think about that ?

### DIFF
--- a/powerdnsadmin/models/user.py
+++ b/powerdnsadmin/models/user.py
@@ -5,6 +5,7 @@ import bcrypt
 import pyotp
 import ldap
 import ldap.filter
+from collections import OrderedDict
 from flask import current_app
 from flask_login import AnonymousUserMixin
 from sqlalchemy import orm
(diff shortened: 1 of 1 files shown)


jbe-dw:
Hello,

I agree that it behaves like you described. But i'd prefer to do extra requests as a fallback:

- We run the initial request to find the user
- If there's an answer, it means that the user is in one group or subgroup
- Then you can test if any of the 3 groups is in memberOf attribute
- If not, do extra requests to requests until matching the highest group

You can add extra elif clauses instead of replacing them, and also only set the role inside all these tests and have a single declaration for the logger, but after the final else clause if no match is done.

I think that the user_ad_member_of tests are not working as we we should not arrive in the final else clause in this case, but it's late. Can you check it as well ?

Thank you for submitting !

melck:
Hello 

I finally found a solution to avoid performing multiple queries. The current code is really hard to read.

I was thinking of exporting all LDAP related queries to `powerdns admin/lib/ldap utils.py` files and when possible reuse connection.

What do you think about that ?

Do you prefer a update only with solution for this issue or can i integrate files changes for better readbility ?

jbe-dw:
Hello,

Can you push updated code as-is first ? I'll take extra time to check the whole code around in the weekend and give you an answer about exporting the code to a library.

Thank you

jbe-dw:
There is no coding rules here so the it is made of different portions written by users with different experience in programming. I'm considering hardening this in a later version, if it happens someday, built from a shiny new blank repository (but the road to this is a hell of a long way).

melck:
@jbe-dw I hope i understand your review. I did otherwise. I am doing a search for available security groups of which the user is a nested member. Then I look at the presence of the groups by importance of roles. This way we make a single request.

Regarding testing, i test my changes on a microsoft active directory. I didn't see unit testing about ad but let me know if miss something.

Regarding the style of code, I noticed it was very different from place to place. For my part on my python projects, I use black to avoid to much different coding styles. Let me know if you prefer that i replace double with single quotes, etc.

melck:
@jbe-dw If you are interested in moving ldap code outside of the user model. I pushed a branch on my repository where I did this work. It's not fully tested but it gives an idea.

https://github.com/melck/PowerDNS-Admin/tree/move-ldap-to-lib

jbe-dw:
The other maintainers are not back from summer break yet, so let's keep the in code version and save your proposal for the v2 with later refactoring.

I'll take some extra time to test it here before validating this one, but the code looks ok to me.

melck:
Thanks @jbe-dw,

I'll be happy to contribute if you want help for a v2. Do you have a issue on that ?

jbe-dw:
Thank you for the proposal. I'm alone for now and quite busy with structural changes in my company. Our plan was to get a fixed version of PDA before trying to work on v2.

Sometimes I'm eager to start it to work on a fresh clean codebase but working on it on my own is far from realistic. I'm thinking about the new architecture (not quite far from what it is now, still based on flask/SQLa). There will be a core application with a rethought API, handling almost all existing features but also multiple backend / authentication backends support, and a separate GUI dialoging with the API.

I'll ping you when it will start but honestly I can't tell you when (you could have moved on to something else then or stopped using PDA). 
For now I'll take the necessary time to check and merge your code as soon as I can.

Regards

SPlusJ:
Any eta on when this will be implemented? Currently using an older version and cannot update until this is fixed.

nkukard:
@AzorianMatt does this one need to be rebased off dev? I happen to have a nested LDAP directory I can test out on if testing is still needed.

AzorianMatt:
@nkukard it needs rebased and properly tested to ensure that it doesn't break the current functionality of the feature. I wish I knew more to tell you on this, but it's another feature I'm not intimately familiar with. If you have the capacity and environment to do so, it would be very much appreciated!

nkukard:
> @nkukard it needs rebased and properly tested to ensure that it doesn't break the current functionality of the feature. I wish I knew more to tell you on this, but it's another feature I'm not intimately familiar with. If you have the capacity and environment to do so, it would be very much appreciated!

No probs, I'll check it out.

nkukard:
I'll be testing this hopefully in the next few hours, terribly sorry for the delay :)

nkukard:
Tested and confirmed working with nested groups. +1 for merge from me.

PR #1246, opened 2022-09-06
admin_edit_key: default to User role for new api keys

unilogicbv:
hopefully this will prevent accidental administator api keys from being created


jbe-dw:
Agreed.

AzorianSolutions (approved):

PR #1247, opened 2022-09-06
models/user.py: properly guard plain_text_password property

unilogicbv:
Resolves the following issue, which occurs with force_otp enabled
and OAuth authentication sources:

File "/srv/powerdnsadmin/powerdnsadmin/models/user.py", line 481, in update_profile
  "utf-8") if self.plain_text_password else user.password
AttributeError: 'User' object has no attribute 'plain_text_password'


AzorianSolutions (approved):
Looks good.

Issue #1265, opened 2022-10-10
Notifications with ACME-Lego (v 4.9.0) not proxyfied through powerdns-admin


WhatshallIbreaktoday:
Hello all, 

After testing last release of ACME-Lego release (which introduces asking to server to send notifys to secondaries servers, I ended up on the following behaviour :  {"msg":"Bad Request"}

reproduced the issue with curl:

**With powerdns-admin proxifying  api request to pdns :** 

curl -k --location --request PUT 'https://myserverip/api/v1/servers/localhost/zones/mytestzone/notify' --header 'Content-Type: application/json' --header 'X-API-Key: mysecretapikey'

Response:
{"msg":"Bad Request"}

**ACME-Lego directly to pdns API :** 

curl -k --location --request PUT 'https://myserverip/api/v1/servers/localhost/zones/mytestzone/notify' --header 'Content-Type: application/json' --header 'X-API-Key: mysecretapikey'
Response:
{"result": "Notification queued"}

After looking at code : [https://github.com/go-acme/lego/pull/1679/files](https://github.com/go-acme/lego/pull/1679/files)
It seems the request is sent with empty body (as also described in  [pdns documentation](https://doc.powerdns.com/authoritative/http-api/zone.html#put--servers-server_id-zones-zone_id-notify)

Workaround : stick to previous ACME-LEGO version, or  add  '{ }' as body (which is also supported by pdns  by the way...)

I'm not able to propose a PR to fix this.. and also wanted to discuss the issue with you before attempting anything...

Best regards,

jbe-dw:
Hello,

All requests sent to api/v1/servers/localhost are forwarded to PowerDNS directly. You should check on PowerDNS API which format is required (and hence do your test directly on PowerDNS API ~ ie not through PDA).

If you would like that PDA silently handles the empty body for LEGO, you can open a PR to add a decorator that fills the body content when sending to notify.

WhatshallIbreaktoday:
Hello,   found the issue, and how to solve it, PR incoming:

For the details : The Bad Request is performed by this code : 
`request.get_json(force=True)`  in helper.py

Following  this post [https://stackoverflow.com/questions/39451002/ensure-the-post-data-is-valid-json](https://stackoverflow.com/questions/39451002/ensure-the-post-data-is-valid-json)  (second answer) and the Flask documentation, 

adding silent=True solves the issue. 
I can now have the same response on the notify request than the native pdns API one.

[...]

PR #1266, opened 2022-10-12
allow null/None JSON data (Used for pdns notifies via api and by LEGO-ACME v 4.9.0)

WhatshallIbreaktoday:
This change permits to proxy pdns zone notify api requests (which are expected to be with empty body)

[pdns api reference](https://doc.powerdns.com/authoritative/http-api/zone.html#put--servers-server_id-zones-zone_id-notify)

Solves #1265


AzorianSolutions (approved):
This seems okay since the app's exception handling isn't very robust.

PR #1301, opened 2022-12-09
enhancement: Small fix

pneb:
I added the parentheses to the `db.session.rollback` line to call the method, which will now properly roll back any changes made to the database if an error occurs.


AzorianSolutions (approved):
Looks good.

PR #1322, opened 2022-12-12
Fixing Validation problem on LDAP form

### LINKED ISSUE #1208: Ldap Authentication validation bug
Filling the Active Directory page does not enable the "save" button.

I had to switch back to LDAP, file in all fields, click save, the group filters are requested even if i say no (toggle on and off solves it), then save, and switch back to Active Directory

Metrax:
Fixes #1208


jbe-dw:
I think I have tested these changes locally when I opened the bug but it is not enough to make all buttons functional.

Have to test on a blank env.

Issue #1306, opened 2022-12-09
module 'ldap' has no attribute 'OPT_X_TLS' after upgrade


mastermueller:
Hi!
Just updated my docker container today, now LDAP Login throws the error 
_module 'ldap' has no attribute 'OPT_X_TLS'_
The LDAP login worked before the upgrade.
I can still login with LOCAL auth, LDAP-Settings seem fine to me.

pneb:
Would you mind pasting the whole traceback error here?

jbe-dw:
Hello,

This is due to a patch upgrading to python-ldap 3.4.2, this version removed this deprecated option. We will release an update quickly.

jbe-dw:
@mastermueller Can you try to remove the line targeted by this PR #1323 ?

You can do it by editing the file in the app folder of the container in an interactive shell. You must restart the container to load the change.

mastermueller:
Hi!
I can confirm this fix does the trick, thank you!

 

Von: jbe-dw ***@***.***> 
Gesendet: Montag, 12. Dezember 2022 16:03
An: PowerDNS-Admin/PowerDNS-Admin ***@***.***>
Cc: mastermueller ***@***.***>; Mention ***@***.***>
Betreff: Re: [PowerDNS-Admin/PowerDNS-Admin] module 'ldap' has no attribute 'OPT_X_TLS' after upgrade (Issue #1306)

 

@mastermueller <https://github.com/mastermueller>  Can you try to remove the line targeted by this PR #1323 <https://github.com/PowerDNS-Admin/PowerDNS-Admin/pull/1323>  ?

You can do it by editing the file in the app folder of the container in an interactive shell. You must restart the container to load the change.

AzorianSolutions:
@mastermueller thank you for pointing this out and my apologies for any complications caused. The current project's stability is low so it's an expected pain unfortunately.

PR #1323, opened 2022-12-12
Draft: Fix pyhton-ldap upgrade

jbe-dw:
This should work. It needs to be tested on a StartTLS setup as well.

I guess that these settings aren't even applied, as the [documentation says](https://www.python-ldap.org/en/python-ldap-3.4.2/reference/ldap.html#tls-options) that we have to call ldap.OPT_X_TLS_NEWCTX to make all the configurations effective, and it is not the case apparently.

At least, if removing this setting is harmless, it could be a quick fixup.


jbe-dw:
Closes #1306

AzorianSolutions:
@jbe-dw thank you for the fast turn around on this!

PR #1328, opened 2022-12-13
Update login.html

Ssshafi:
Use SITE_NAME for login box title on login page.


Metrax (approved):
LGTM

PR #1339, opened 2022-12-22
Improve bugfix tests

ymage:
Heavily based on #1198 from @corubba (Many thanks)


AzorianMatt:
@ymage thank you for your additional work on this! From what I can tell, these changes look good but I need to pull in another contributor for review before we accept the changes.

PR #1343, opened 2022-12-23
Bump alpine to 3.17 in dockerfile

trappiz:
Alpine 3.13 is EOL, quick fix to get powerdns-admin running on alpine 3.15 :)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.15 AS builder
+FROM alpine:3.17 AS builder
 
 ARG BUILD_DEPENDENCIES="build-base \
     libffi-dev \
@@ -29,7 +29,7 @@ COPY ./requirements.txt /build/requirements.txt
 
 # Get application dependencies
 RUN pip install --upgrade pip && \
-    pip install -r requirements.txt
+    pip install --use-pep517 -r requirements.txt
 
 # Add sources
 COPY . /build
@@ -65,7 +65,7 @@ RUN mkdir -p /app && \
     cp -r /build/configs/docker_config.py /app/configs
 
 # Build image
-FROM alpine:3.15
+FROM alpine:3.17
 
 ENV FLASK_APP=/app/powerdnsadmin/__init__.py \
     USER=pda
@@ -79,7 +79,7 @@ RUN apk add --no-cache mariadb-connector-c postgresql-client py3-gunicorn py3-py
     apk del libcap
 
 COPY --from=builder /usr/bin/flask /usr/bin/
-COPY --from=builder /usr/lib/python3.9/site-packages /usr/lib/python3.9/site-packages/
+COPY --from=builder /usr/lib/python3.10/site-packages /usr/lib/python3.10/site-packages/
 COPY --from=builder --chown=root:${USER} /app /app/
 COPY ./docker/entrypoint.sh /usr/bin/
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.13 AS builder
+FROM alpine:3.15 AS builder
 
 ARG BUILD_DEPENDENCIES="build-base \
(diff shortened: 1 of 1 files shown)


AzorianMatt:
@trappiz I noticed the addition of these two Python modules: py3-pyldap, py3-flask

They seem like obvious additions that I question why they were missing in the first place but no one has said anything that I'm aware of. Can you elaborate on what lead to this particular part of the change?

christianbur:
Maybe we should consider an update to Alpine 3.17 right away.

AzorianMatt:
@christianbur seems reasonable if the effort is going to be put forth anyhow.

trappiz:
Had it in mind but didn't want to bump to many versions of Alpine right away due to package versions etc.

I will look into building using 3.17 👍

-------- Original Message --------

AzorianMatt:
@trappiz thank you!

christianbur:
I just tried it myself with alpine:3.17, unfortunately so far without success.

Changes in Dockerfile:
 - "FROM alpine:3.13 AS builder" -> "FROM alpine:3.17 AS builder"
- "FROM alpine:3.13" -> "FROM alpine:3.17"
- "COPY --from=builder /usr/lib/python3.8/site-packages /usr/lib/python3.8/site-packages/" -> COPY --from=builder /usr/lib/python3.10/site-packages /usr/lib/python3.10/site-packages/

The image creation was successful, but the log shows the following error:
`ImportError: Error loading shared library libldap.so.2: No such file or directory (needed by /usr/lib/python3.10/site-packages/_ldap.cpython-310-x86_64-linux-gnu.so)`

I could not find the file libldap.so.2 with the 'find' command in the whole image.

jbe-dw:
`apk add libldap` ?

trappiz:
These packages needs to be added on row 73.

py3-pyldap py3-flask

I will update this PR to 3.17 this afternoon when I'm back home again :)

-------- Original Message --------

docker/Dockerfile
```diff
@@ -29,7 +29,7 @@ COPY ./requirements.txt /build/requirements.txt
 
 # Get application dependencies
 RUN pip install --upgrade pip && \
-    pip install -r requirements.txt
+    pip install --use-pep517 -r requirements.txt
```
trappiz, line 32: Added --use-pep517 to get ride of deprecation warnings.

> DEPRECATION: rjsmin is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559

trappiz:
> @trappiz I noticed the addition of these two Python modules: py3-pyldap, py3-flask
> 
> They seem like obvious additions that I question why they were missing in the first place but no one has said anything that I'm aware of. Can you elaborate on what lead to this particular part of the change?

Reason for these additions were that before when starting the container it complained about missing stuff :(

I'm no python dep expert so maybe it could it be resolved by updating some python dep?

trappiz:
@AzorianMatt @christianbur please have a look at the latest commit. Building the image works fine, started it up quickly and was able to create a user, login etc :smiley:

christianbur:
Everything looks good for me too :-)

- Images built
- Logged in with 2FA
- Created new DNS entry, tested and deleted
- Docker log is also OK.

AzorianMatt:
@trappiz @christianbur first off, thank you both for your work on this! Now that we have a "working" build, it still must be personally tested by someone using the third-party authentication integrations to ensure that I don't roll another broken image out again (or at least make a good effort not to).

@jbe-dw are you able to test this by chance?

jbe-dw:
I can test it next week

jbe-dw:
I can successfully login with LDAP with this image.

AzorianMatt:
Thank you very much @jbe-dw! I'll merge it in now.

PR #1357, opened 2023-01-11
fix: Potential fix for a regex bug

### LINKED ISSUE #1338: [Bug] Error when login with AD groups
Hi,
my company uses PowerDNS-Admin and noticed that after an update some users experience app crashes when login.
It applies only to users matched by regex for AD security groups.

We found the UnboundLocalError for "account" variable ([line](https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/0404748e6e71851f28bc56f7ea2983eac83fac19/powerdnsadmin/routes/index.py#L327))
 `account_id = account.get_id_by_name(account_name=sanitized_group_name)`

in index.py after https://github.com/PowerDNS-Admin/PowerDNS-Admin/commit/a87b9315202fbe6af50293874df094ccde2a54e3 update and had to re-initialize it there.

pneb:
Might fix #1338 

Feel free to revert the changes if not

### DIFF
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -323,6 +323,7 @@ def login():
                             # Regexp didn't match, continue to next iteration
                             continue
 
+                    account = Account()
                     sanitized_group_name = Account.sanitize_name(group_name)
                     account_id = account.get_id_by_name(account_name=sanitized_group_name)
 


AzorianMatt:
@pneb I'm really unclear on the intention with this particular change? The best I can tell, you added a single line that instantiates an instance of the Account object. Looking at the code to follow before the return statement, both paths set their own instance of this variable so it would appear it has no effect on anything.

Can you elaborate on this? Surely I'm just missing something right?

khayalhus:
I was trying to integrate Microsoft OAuth into PowerDNS and stumbled upon this PR. As described in this issue https://github.com/PowerDNS-Admin/PowerDNS-Admin/issues/1338 , when a user tries to sign up through Microsoft, the error happens during the process of synchronizing the groups the user is a member of with the PowerDNS-Admin's Account feature. "account" variable is called before initialization. [Looking at the codebase, ](https://github.com/PowerDNS-Admin/PowerDNS-Admin/search?q=get_id_by_name) a similar approach of instantiating and then discarding an object has been taken a few more times to query the database. This PR adds back a line that has been removed in a previous commit referenced in the issue.

Nevertheless, this line of code has solved the issue for me.

PR #1365, opened 2023-01-25
Update base.html

Ssshafi:
Same as PR #1328
Use SITE_NAME for upper left title on base page.

### DIFF
--- a/powerdnsadmin/templates/base.html
+++ b/powerdnsadmin/templates/base.html
@@ -32,7 +32,7 @@
       <!-- mini logo for sidebar mini 50x50 pixels -->
       <span class="logo-mini"><b>PD</b>A</span>
       <!-- logo for regular state and mobile devices -->
-      <span class="logo-lg"><b>PowerDNS</b>-Admin</span>
+      <span class="logo-lg"><b>{{ SITE_NAME }}</span>
     </a>
     <!-- Header Navbar: style can be found in header.less -->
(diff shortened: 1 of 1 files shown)


powerdnsadmin/templates/base.html
```diff
@@ -32,7 +32,7 @@
       <!-- mini logo for sidebar mini 50x50 pixels -->
       <span class="logo-mini"><b>PD</b>A</span>
       <!-- logo for regular state and mobile devices -->
-      <span class="logo-lg"><b>PowerDNS</b>-Admin</span>
+      <span class="logo-lg"><b>{{ SITE_NAME }}</span>
```
Metrax, line 35: Hi,

thank you for you PR.

Unfortunately, this will break the HTML markup, because you are opening the <b>, but you are not closing it.

Additionally you should make sure, that if SITE_NAME is the default value "PowerDNS-Admin" the current markup will persist.

Please adapt your PR, and we can merge it into the PDA code.

Bests,
Robert

Metrax (changes requested):

powerdnsadmin/templates/base.html
Ssshafi: Sorry for the mess ...
I made a new PR without error, with default value.
The changes concern base.html (#1365 ) and also login.html (#1328 )

Metrax:
Contributor opened a new PR #1366

PR #1366, opened 2023-01-25
Use SITE_NAME as title

Ssshafi:
Use SITE_NAME for upper left title on base page and login box title on login page (with default value).


AzorianMatt:
@Ssshafi @Metrax thank you both for your work on this quick tweak. Definitely a welcome change that makes sense.

PR #1370, opened 2023-02-02
Force a random SECRET_KEY on boot

Peyphour:
This default key should not exists.

### DIFF
--- a/powerdnsadmin/default_config.py
+++ b/powerdnsadmin/default_config.py
@@ -4,7 +4,7 @@ basedir = os.path.abspath(os.path.dirname(__file__))
 
 ### BASIC APP CONFIG
 SALT = '$2b$12$yLUMTIfl21FKJQpTkRQXCu'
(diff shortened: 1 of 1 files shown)


AzorianMatt:
@Metrax @jbe-dw are either of you able to verify my findings that the app's SECRET_KEY config option is only currently used for URL serialization / protection? I want to ensure I haven't made an oversight in assuming that this key is **not** currently being used for symmetric encryption anywhere in the app.

Peyphour:
> 

Hey, actually it is being used to sign session cookies. It is possible to bypass authentication for deployments using the default key

AzorianMatt:
@Peyphour thank you for catching that. For that reason, I am rejecting this change as it will break horizontally scaled deployments as different secret keys would be generated for each instance of the application running, thus creating issues with non-sticky, load-balanced sessions.

I agree that default keys are bad, but then again so is not following the deployment guide. I'm of the mentality that operators are ultimately responsible for mitigating these risks. I'm all for a change that solves this without breaking the aforementioned use-case.

PR #1371, opened 2023-02-06
[WIP] Admin LTE & Bootstrap Upgrade

AgentTNT:
This would replace https://github.com/PowerDNS-Admin/PowerDNS-Admin/pull/1173

Convert col-xs-* to just col-* as part of bootstrap v3 -> v4
Convert box-* -> card-* as part of bootstrap v3 -> v4
Font Awesome v4 -> v5 (can go to v6 but small iterative updates first)
Moved domain actions on main dashboard to a dropdown menu to avoid clutter in rows
Added "Log Out" to top header left
Hid OTP on admin edit user to only show the disable card & options if the user account has OTP enabled
Certain modals were updated to fix positioning of the "X" within the header

login page has red card header to denote testing/danger and to not be used with prod systems right now

"flask assets build" takes a bit - will need to work on trimming this down
awesome fonts was moved from packages to be loaded in the CDN, but will convert back to packages when editing is done


AgentTNT:
Icons in buttons also started transition from right of text to left of text. Some buttons lack an icon at the moment, but will be resolved. 

image
image
image
image

AzorianMatt:
@AgentTNT love the work you're doing, it is very appreciated to say the least! I'll be a little slow at getting merges done, but once I can test them, I'll get them through for you. Thank you!

AzorianMatt:
@AgentTNT here's what I have so far, more review to come:

- On the domain record page, let's keep the "Apply Changes" button aligned to the far right as it has been. I can see that this particular change would serve as an annoyance to those that are used to a speedy workflow using the old position.

So far, most things look great for the change over (as compared to their prior versions). I have many styling adjustments I want to make, but I'll save those for my pass after this is finished and merged.

PR #1377, opened 2023-02-13
API record change history format fix

### LINKED ISSUE #1375: Per-record changelog does not show changes made through API
When looking at the history of a domain record (Dashboard -> Select Domain -> Click on the clock icon of a record), only changes made by Accounts from the web UI are present.

I believe this is because actions made from UI are logged with "add_rrsets" and "del_rrsets" keys, while actions from the API are logged as the PowerDNS Authoritative API format which is "rrsets". Below are the database entries for API and UI, respectively.

```sql
id         | 248
msg        | replace zone A record of test.example.com
detail     | {"rrsets": [{"name": "test.example.com.", "type": "A", "ttl": 86400, "changetype": "REPLACE", "records": [{"content": "10.25.25.29", "disabled": false}]}]}
created_by | api-test
created_on | 2023-02-08 12:13:08.229078
domain_id  | 3
-[ RECORD 234 ]-
id         | 250
msg        | Apply record changes to domain example.com
detail     | {"domain": "example.com", "add_rrsets": [{"name": "test.example.com.", "type": "A", "ttl": 86400, "records": [{"content": "10.25.25.30", "disabled": false}], "comments": [{"content": "", "account": ""}], "changetype": "REPLACE"
}], "del_rrsets": [{"comments": [{"content": "", "account": ""}], "name": "test.example.com.", "records": [{"content": "10.25.25.29", "disabled": false}], "ttl": 86400, "type": "A", "changetype": "DELETE"}]}
created_by | khayalhus@example.com
created_on | 2023-02-08 12:28:38.131618
domain_id  | 3
-[ RECORD 236 ]
```

Looks like the following function is responsible for this display. When retrieving logs per record, it only looks for the "add_rrsets" and "del_rrsets" keys.

https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/c5b9e2460498e28baad214afc8c9217a25d1f4d5/powerdnsadmin/routes/admin.py#L906

My Setup:
- Environment: _Docker_
- PowerDNS-Admin image version: _powerdnsadmin/pda-legacy:latest_
- PowerDNS-Admin backend: _postgres:15.1_
- PowerDNS-Authoritative image version: _powerdns/pdns-auth-47:latest_

raunz:
Resolves #1375 by changing record change log format to match generic logging format.
With API patch changing multiple records, also a single logging entry is created according to the change (removed unnecessary for loop).


AgentTNT:
Something with this breaks if there is no description for the key OR something else, when using terraform and API keys to use a 
DNS challenge on a API key bound to a domain (did not test with the account association). It would render the history function inoperable and fault the history page from loading the tab contents  and same with the dashboard page if it was within the last 5 events.

image

raunz:
> Something with this breaks if there is no description for the key OR something else, when using terraform and API keys to use a DNS challenge on a API key bound to a domain (did not test with the account association). It would render the history function inoperable and fault the history page from loading the tab contents and same with the dashboard page if it was within the last 5 events.

@AgentTNT - could You share the exact history entry from SQL? 

I've added histories with and without:  "comments": [{"content": "", "account": ""}]  key/value sets and can't freak the UI out. e.g.:
{"domain": "example.com", "add_rrsets": [{"name": "_acme-challenge.example.com.", "type": "TXT", "ttl": 600, "changetype": "REPLACE", "records": [{"content": "\"asdasdasdasdasd\"", "disabled": false}]}], "del_rrsets": []}

AzorianMatt:
@raunz @AgentTNT have you two been able to work out if this is indeed a breaking change or not?

@raunz it seems there are some merge conflicts that will need to be resolved. By later today, I hope to have a big set of changes from @AgentTNT merged in which will give a more appropriate base to work this from.

PR #1378, opened 2023-02-13
Global Search available for all users

raunz:
Introducing Global Search feature for standard users too. As the API search call has no restriction, the results are filtered by allowed domain list for standard users. If standard user has a lot of domains to manage, this feature is a must have.
Operators/administrators search results won't be filtered.

### DIFF
--- a/powerdnsadmin/routes/admin.py
+++ b/powerdnsadmin/routes/admin.py
@@ -2021,7 +2021,6 @@ def delete_template(template):
 
 @admin_bp.route('/global-search', methods=['GET'])
 @login_required
-@operator_role_required
 def global_search():
     if request.method == 'GET':
         domains = []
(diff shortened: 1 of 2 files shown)


raunz:
@AzorianMatt - I tried out the 0.4.0 release and it seems, that part of my PR powerdnsadmin/templates/base.html got accidentally reverted by one of your [commits](https://github.com/PowerDNS-Admin/PowerDNS-Admin/commit/26f3f79388a5e8587d9e8f704cc868398e022c94#diff-ce29ab81a9416b0a7fbbb301873b4d3d6932ad2e8b3b0ac5579a02eb3b52c560)
Standard user can access /admin/global-search , but the menu item isn't available. 
Screenshot_20230312_160720

AzorianMatt:
@raunz We can re-introduce this as needed but I'm guessing the part that got reverted was the "Global Search" menu option that landed right in the middle of the `Zone Management` section? If so, that's definitely not an acceptable change as it made no sense (IMO) to be where it was.

Was the intent here to keep it outside of the `Administration` section of the side navigation? Please advise and I'll make the change necessary to get it placed properly in the navigation.

AzorianMatt:
@raunz I just pushed a change to the dev branch that should take care of this.

raunz:
> @raunz I just pushed a change to the `dev` branch that should take care of this.

That looks just excellent in the dev branch. I just looked at the old base.html, that the History related if clause in the end wasn't great and the entire setup for "Administration" part needs some rearrangement - but You already solved it nice and clear. Many thanks! 
Can we have that change in 0.4.1 ?

AzorianMatt:
@raunz thanks and you're welcome! I hadn't intended on changing the activity item during this but had noticed the menu item was duplicated as a poor attempt to deal with the multiple permissions scenario. As a result, it had not been updated to reflect the latest styling changes so this was my quick way of fixing that problem moving forward.

These changes will be adopted into the 0.4.1 release and should remain stable in the `dev` branch until then if you need immediate access to a Docker build until the next release.

PR #1392, opened 2023-02-19
Issue 1391 - Feature: Mobile First Responsive UI Design

AzorianMatt:
Actively working on [issue 1391](https://github.com/PowerDNS-Admin/PowerDNS-Admin/issues/1391) which should hopefully be ready sometime this week.

Updates:

Many changes were made here but most notably, some features were given updated titles such as "History" being changed to "Activity". The server statistics and configuration viewing feature have been split into separate pages. Some menu items were given different icons and titles in specific cases. Most importantly, all views have been brought closer to a somewhat uniform standard as part of a bigger long-term effort toward other clean-up. Some views were re-designed slightly to better accommodate mobile layouts but the impact should be fairly minimal when considering future plans to come.

The activity feature (previously known as history), is not cleaned up as part of this. It will require a fairly hefty re-design as the HTML markup is very bad and completely broken. It should have never been accepted in this state but we'll have to live with it a bit longer until time and motivation permits.


AzorianMatt:
The activity feature (previously known as history), is not cleaned up as part of this. It will require a fairly hefty re-design as the HTML markup is very bad and completely broken. It should have never been accepted in this state but we'll have to live with it a bit longer until time and motivation permits.

PR #1393, opened 2023-02-20
fix: configurable CAPTCHA_ENABLE env when docker use

ichisuke55:
Recently there was a PR merge where CAPTCHA_ENABLE environment variable was enabled by default, but it is easier to use if the user can disable it, so I made it changeable via environment variable.


AzorianMatt:
Great catch @ichisuke55!

PR #1335, opened 2022-12-18
Add OIDC env vars and set SAML_ENABLED as False if unset

ymage:
Fix #1334

PR #1397, opened 2023-02-23
Add support for oidc_oauth_metadata_url configuration option

AdvanticGmbH:
This commit adds support for the `oidc_oauth_metadata_url` configuration option. This option specifies the URL of the OIDC server's metadata endpoint, which contains information about the OIDC server's endpoints, supported scopes, and other configuration details. By using this option, we can ensure compatibility with different OIDC servers and reduce the risk of errors due to manual endpoint configuration.


AdvanticGmbH:
This behaviour can be tested with Auth0. Without configuring the metadata url the authentication wont work.

AdvanticGmbH:
Possible duplicate with #1335

PR #1426, opened 2023-03-08
Small fix for subfolder proxying, Documentation, small fix for build process hangs

ckbaker10:
I had the issue that the build from master was hanging up on "flask assets build"
I switched the the filter to rcssmin, cssmin is only available in 0.2.0 in pip which is stone age old.

rcssmin was added to the requirements, build works fine.

I documented how to set up an apache reverse proxy with subfolder using the SCRIPT_PATH environment variable.

There was a small error in powerdnsadmin/templates/admin_history.html using static urls instead of url_for

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -38,7 +38,7 @@ COPY . /build
 # Prepare assets
 RUN yarn install --pure-lockfile --production && \
     yarn cache clean && \
-    sed -i -r -e "s|'cssmin',\s?'cssrewrite'|'cssmin'|g" /build/powerdnsadmin/assets.py && \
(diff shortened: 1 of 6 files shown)


AzorianMatt:
@ckbaker10 the build process doesn't hang but just takes a really long time currently. Does your replacement notably speed that up?

ckbaker10:
@AzorianMatt I left it running for about 10 minutes and then canceled, with rcssmin the flask assets build process ends within a few seconds.

AzorianMatt:
@ckbaker10 great to hear about such an improvement. I'll test as soon as I can later this week and look at getting that merged in. The current process was taken between 10 to 15 minutes on average for me.

AzorianMatt:
@ckbaker10 This has been tweaked and merged into the newly created "dev" branch which will hold on-going pre-release dev updates.

Issue #1307, opened 2022-12-09
Current Master Branch Pip Install Fails Due To lxml


AzorianSolutions:
I could not build the current master branch on my Ubuntu 22.04 machine due to a conflict arising with lxml 4.8.0. I could not find the source of this dependency using pipdeptree -r -p lxml.

I was able to successfully move past this build issue by manually setting the lxml dependency in requirements.txt for version 4.9.1. PR coming shortly.

PR #1433, opened 2023-03-12
Update lxml version to support python3.11

subbink:
This PR updates lxml version to suport python3.11
python3-saml also has to be updated because of a dependancy.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,7 +25,7 @@ gunicorn==20.1.0
 itsdangerous==2.1.2
 jsonschema[format]>=2.5.1,<4.0.0 # until https://github.com/Yelp/bravado-core/pull/385
 lima==0.5
-lxml==4.6.5
+lxml==4.9.0
 mysqlclient==2.0.1
 passlib==1.7.4
 #pyOpenSSL==22.1.0
@@ -33,7 +33,7 @@ pyasn1==0.4.8
 pyotp==2.8.0
 pytest==7.2.1
 python-ldap==3.4.3
-python3-saml==1.14.0
+python3-saml==1.15.0
 pytimeparse==1.1.8
 pytz==2022.7.1
 qrcode==7.3.1


AzorianMatt:
@subbink we will definitely need some of the other members to review and verify this as I can't help but think I remember there being some issue that previously popped up and lead to a change revert in relation to upgrading the lxml package.

Does this ring a bell @jbe-dw or @corubba?

AzorianMatt:
@corubba thank you. That rings a bell and I believe that is what I was thinking of. It doesn't seem like a great idea for us to proceed with the addition of something that isn't yet officially supported. I realize there are probably other instances like this in the project already, but I'm not trying to make matters worse.

AzorianMatt:
@subbink I am rejecting this PR for the time being as it seems that the required changes would introduce another unsupported configuration into the project. If this can be disproved or a reasonable alternative ensured, then I would be happy to reconsider.

nkukard:
references #1442 

The issue with python3-saml not working is because the binary lxml wheel is built against a different version of libxml2.

``--no-binary lxml`` can be used to fix this and python3-saml will work.

for references check these:
- https://github.com/onelogin/python3-saml/issues/292
- https://bugs.launchpad.net/lxml/+bug/1960668
- https://github.com/open-formulieren/open-forms/pull/2247

Tested SAML and working with Docker container running "3.17" replaced with "edge" with this PR applied.

Please kindly test and reconsider applying with above a adjustment to prevent binary installation.

If your SAML testing fails please provide me with the log/error.

AzorianMatt:
@nkukard this is great information, thank you!

I will reopen and get this done.

nkukard:
I'm busy testing a MR to fix this, will submit soon.

Metrax:
Please don't forget, that we had some issues with high lxml version, see  #1307 

Please check this PR against various distributions!

nkukard:
> Please don't forget, that we had some issues with higher lxml version, see #1307
> 
> Please check this PR against various distributions!

It would of helped if logs were included. I will ask for them.

Is the recommended way of installing PowerDNS Admin not using a venv? or the Docker container?

It is unlikely any distribution will have the versions listed in requirements.txt installed by the OS as packages.

Metrax:
> Is the recommended way of installing PowerDNS Admin not using a venv? or the Docker container?

Locally in a venv and the public docker image are the recommended options. AFAIK we have removed all documentations working without venv.

Some python packages have dependencies with system libraries or system tools. These dependencies can break the installation undepended from venv.

nkukard:
Should be fine. Is there perhaps a list of supported distributions and their versions? I'll add it to my test framework.

subbink:
This PR is no longer needed, because PR #1451 implements the same, but in a more compatible way.

PR #1434, opened 2023-03-12
PostgreSQL Support Improvement

subbink:
This PR contains things to improve using PostgreSQL for PowerDNS-Admin. It contains both documentation changes as well as an addition to the requirements.txt and the docker files which are needed for that.


nkukard:
I'm not a core developer, but this looks good to me and would be a good improvement imho.

PR #1437, opened 2023-03-12
Added customizeabllity for created_by column in history

benshalev849:
I have added an option to create a custom header to be displayed in the history created_by segment.

This is used to allow custom headers to be in the option instead of just the description of the api-key which might not be enough for auditing.

The header is customizable and is a string, the code checks if the setting is not null and if it exists in the request.
If it exists then input it into the history, if not input the normal api-key description.

This change is only to the API.


AzorianMatt:
@benshalev849 I am rejecting this PR as it does not follow the newly implemented contribution requirements outlined in the contribution guide of the project. Please refer to the README.md for a reference to the contribution guide.

Nevertheless, thank you for your efforts.

benshalev849:
@AzorianMatt 

If i merge to `dev` is that okay?

AzorianMatt:
@benshalev849 I would rather hold a discussion on this change with other members to get a better idea of the intent here. I'm not saying it's not a good change, but I fail to understand it's universal application on first review. If the value of this change can be re-enforced by some other users, then it will be reconsidered but as of now, it appears to be a one-off change for your use-case.

Additionally, merging to `dev` instead of `master` is another issue with this PR outside of not starting a discussion about it first.

Thanks!

benshalev849:
@AzorianMatt 
Okay, created the discussion and i hope more people will see the need so it can be merged into the upstream, once i see it is discussed and approved i will create a new request to merge into the `dev` branch :)

AzorianMatt:
I appreciate your participation on the matter @benshalev849! I'll head over to the discussion and add my notes.

AzorianMatt:
@benshalev849 another important note when you re-introduce this PR, you need to make appropriate documentation updates so that others can be aware of it's existence. There's no point in me allowing a new feature if it only exists in code and no documentation.

raunz:
@benshalev849 - I understand the issue You have, but the variable naming in code isn't quite spot on: "custom_header_setting". It would make a lot more sense, if the variable is named something like "changed_by_addition" or something.
Also the actual variable in request should be: "requested_by" or smth. Replacing the apikey.description with custom_header_setting removes all traces to the actual apikey being used - my recommendation would be to add data to the apikey.desc. Then we would have full usable history "Apikey.desc John Doe, special dpt".

Issue #1412, opened 2023-02-28
Unifiy Nomenclature for Zone / Domain References


Metrax:
At some places, we call Zones "Zone(s)", at other places we call it "Domain(s)". We should unify this to "Zone(s)"

AzorianMatt:
@Metrax agreed. I wasn't sure if I got all of the places yet or not but I wasn't very thorough on my efforts during the change. I do want to unify on the appropriate nomenclature of zone though instead of the slang domain.

AzorianMatt:
This has been completed by [PR 1441](https://github.com/PowerDNS-Admin/PowerDNS-Admin/pull/1441).

PR #1441, opened 2023-03-12
Finish Updating Zone Nomenclature From Domain

subbink:
This PR should (mostly) fix issue #1412 by changing the text domain to zone.
I have tested the basic functionality. It would be nice if someone else can also check nothing got broken.

### DIFF
--- a/docs/wiki/README.md
+++ b/docs/wiki/README.md
@@ -38,7 +38,7 @@
 
 ## Using PowerDNS-Admin
 
-- Setting up a domain
+- Setting up a zone
 - Adding a record
 - <whatever else>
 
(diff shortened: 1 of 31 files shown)


AzorianMatt (commented):
@subbink these changes generally speaking look great!

However, I do you see one problem that would be introduced by this change which we will need to tweak the PR for before merging. Due to how the history / activity feature has been written through the use of SQL like queries in some places, the logic will need updated to include not only the new nomenclature (`zone`), but also the previous nomenclature (`domain`). Otherwise, existing history / activity data will not be properly matched by the current logic after this change.

Let me know if you are comfortable working on this particular problem or if I should make the changes.

Thanks!

subbink:
@AzorianMatt, I will have a go to fix those items.

subbink:
I have made some changes to keep the old history items fully functional.

AzorianMatt:
@subbink I'll try to carve out sometime tomorrow to circle back and do reviews and merges. Thank you for addressing that!

Issue #1237, opened 2022-07-22
Feature: configurable OTP label


upcFrost:
Hi,

Right now the OTP label is hardcoded to be `PowerDNS-Admin`
https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/204c996c81f2f2ef14f6d98e5ea907e69a7ede92/powerdnsadmin/models/user.py#L92

It would be great for it to either be configurable or to match the `site_name`, as some authenticator apps tend to overwrite the key if the label is the same. Proposed change:
```diff
--- original.py	2022-07-22 09:08:18.460801206 +0000
+++ user.py	2022-07-22 09:06:39.660791555 +0000
@@ -89,8 +89,8 @@
         return '<User {0}>'.format(self.username)

     def get_totp_uri(self):
-        return "otpauth://totp/PowerDNS-Admin:{0}?secret={1}&issuer=PowerDNS-Admin".format(
-            self.username, self.otp_secret)
+        return "otpauth://totp/{2}:{0}?secret={1}&issuer={2}".format(
+            self.username, self.otp_secret, Setting().get('site_name'))

     def verify_totp(self, token):
         totp = pyotp.TOTP(self.otp_secret)
```

subbink:
It seems to me that this issue can be closed.

AzorianMatt:
@upcFrost this has been resolved in a recent PR and will be included in the next patch release very soon.

Thank you!

PR #1444, opened 2023-03-13
Make the OTP label the same as the site_name #1237

subbink:
This PR solves issue #1237 .


AzorianMatt:
@subbink thank you for completing this!

PR #1457, opened 2023-03-17
Implement password strength & complexity checking

### LINKED ISSUE #989: Local Auth Password Complexity Requirements
When registering an account, there are no password requirements, even length.

For example, one can create a password with a single character. This seems dangerous since self registration is available.

Unsure how exactly to fix, but perhaps config options for

minimum_password_length (int)
valid_password_chars (string?)

and use both to compare, the string could be `abc..ABC..!@#$` etc.

nkukard:
Closes #989

I completely re-implemented https://github.com/PowerDNS-Admin/PowerDNS-Admin/pull/1054, which can be closed if this is merged. I did keep the values that were used as defaults.

Password policies are **opt-in** so we don't affect current installations.

I added password strength and complexity checking in a way it allows you to pick either or both of the methods, namely password character requirement counts and password complexity based on zxcvbn.

If either method is active, restrictions on using the username, password, email address, first name, last name will be set in place. I'm sure we can all agree unanimously on that!

I added help to the right hand side for the "Basic" tab in the "Authentication" settings to help with the fields being entered in.

Blanking the any password policy setting will return it to its default value.

I added password policy checking to user registration and user profile page, as far as I can see this is the only two places the password can be set.

When using a password which does not meet the policy defined in settings, the policy will be displayed along with a `*` for each item which fails to assist in indicating what must be fixed. I think its a good start, I couldn't think of an easier more appealing way to indicate failure on a per-requirement basis.

Only one requirement added to requirements.txt, which is `zxcvnb`.

The changes are pretty simple and straight forward, I tried to keep the patch as clean as possible to assist with review.

I tested each of the password policies, enabling/disabling each method along with testing each policy is honored.

I'd appreciate any testing from the community, but I'd say its ready for merge and if anyone has any issue with the functionality I'll gladly attend.

```
zxcvnb requirement tested on:
    "ubuntu:23.04"
    "ubuntu:22.10"
    "ubuntu:22.04"
    "ubuntu:20.04"
    "debian:10"
    "debian:11"
    "alpine:edge"
    "alpine:3.17"
    "alpine:3.16"
    "alpine:3.15"
    "alpine:3.14"
    "rockylinux:9"
    "almalinux:9"
    "fedora:36"
    "fedora:37"
    "fedora:38"
    "fedora:39"
```

pass3
pass1
pass2

### DIFF
--- a/powerdnsadmin/models/setting.py
+++ b/powerdnsadmin/models/setting.py
@@ -205,6 +205,14 @@ class Setting(db.Model):
         'deny_domain_override': False,
         'account_name_extra_chars': False,
         'gravatar_enabled': False,
+        'pwd_enforce_characters': False,
+        'pwd_min_len': 10,
+        'pwd_min_lowercase': 3,
+        'pwd_min_uppercase': 2,
+        'pwd_min_digits': 2,
+        'pwd_min_special': 1,
+        'pwd_enforce_complexity': False,
+        'pwd_min_complexity': 11
     }
 
     def __init__(self, id=None, name=None, value=None):
--- a/powerdnsadmin/routes/admin.py
+++ b/powerdnsadmin/routes/admin.py
@@ -1550,7 +1550,23 @@ def setting_authentication():
             local_db_enabled = True if request.form.get(
                 'local_db_enabled') else False
             signup_enabled = True if request.form.get(
-                'signup_enabled', ) else False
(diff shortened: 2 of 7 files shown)


AzorianMatt:
Thank you once again @nkukard for your work on this! It's a very welcome addition to the app to say the least.

Issue #1480, opened 2023-03-23
Activity search form doesn't execute - Search button not submitting


raunz:
### PDA version

0.4.0

### Python version

3.0

### Steps to Reproduce

1. Navigate to "Activity" menu.
2. enter any search filter (e.g. "Domain name" in "Search By Domain")
3. press "Search".
4. request won't be sent and results don't change.

Issue is present in version **0.4.0** and also current pda-legacy:**dev** image.

As javascript/html isn't my cup of tea, could someone provide a fix for this.

I would in general like to see, when navigating to "Activity" page, no activity records would be pulled from database automatically. As I have 450K record in history table, any unfiltered query generates a lot of payload. Setting the **max_history_records** number to small value isn't an option, as for some domains I really want to see full history with thousands of activity records. 

### Expected Behavior

Activity output of query with submitted filters.

### Observed Behavior

https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/master/powerdnsadmin/templates/admin_history.html#L482 only fills out the "Maximum date:" field.

By inspection html, there's an issue with 
`<form id="history-search-form" autocomplete="off">...</form>`
and https://github.com/PowerDNS-Admin/PowerDNS-Admin/blob/master/powerdnsadmin/templates/admin_history.html#L487 as the `$("#history-search-form").submit `is outside of the "form".

To get desired results I can execute in browsers console command: `$("#history-search-form").submit()`
or copy a submit button into `<form id="history-search-form"></form> `section.

Sample:
upper "Search" button works
lower "Search" button doesn't
![image](https://user-images.githubusercontent.com/15248710/227196179-476cdd22-cd3e-4d3a-93e1-74d21cca71fc.png)

AzorianMatt:
@raunz please provide your operating system and browser environment information. This feature works just fine for me in all of the mentioned versions that you said it doesn't so I suspect an environmental issue.

Granted, I haven't been cross-browser testing this so my information is based on Chrome.

raunz:
@AzorianMatt - debian11. tried different browsers - behaviour is exactly the same as described above.
Chromium 110.0.5481.177    Firefox 102.9.0esr
I also checked a Windows enviroment Chrome - same problem.

AzorianMatt:
@raunz after another look, here's what I have for you. I **can't** reproduce this in version `0.3.0` but I can reproduce this in `0.4.0`.

[...]

PR #1481, opened 2023-03-25
Fix activity search invalid form markup causing submission failures

raunz:
### Fixes: #1480

I looked at powerdnsadmin/templates/admin_history.html and the search form begins at L43 and should end with `</form>` at L204, but there's in an excessive `</form>` in between at L147.

As the `<form>` and `</form>` lines are in different div's, I moved `<form>` one level up and both div's building the search form are under single form entity. Tested with current dev release.


AzorianMatt:
Thank you for your work on this @raunz!

PR #1520, opened 2023-04-12
Added support for building `ARM` and `ARM64` based Docker images.

AzorianMatt:


AzorianMatt:
The `ARM` support was removed in favor of keeping only `ARM64` after observing the effect on build times.

PR #1580, opened 2023-05-24
Add LDAP to Wiki

kolonuk:
Added LDAP to wiki.

Even as a base, it's better than nothing and I had to really fight hard to find these details.


AzorianMatt:
@kolonuk while I appreciate the efforts, this was not previously discussed and approved per the contribution requirements.

If I don't hold you as accountable as I do the contribution team, then what kind of message would that send?

PR #1602, opened 2023-06-25
(WIP) 1601 - Refactor Domain routes

### LINKED ISSUE #1601: Refactor Domain/Zone Code
### Proposed Changes

Decoupling the domain/zone pages and routes will ease the burden for future coders on current code base.

Currently functions were moved into an AJAX call on page load, and then the rendering will complete on the client side after the JSON payload is acquired. Moving forward with server side CSRF should also help us here. This will also allow the app the be broken out to a frontend and backend separate deployments like a microservice if we choose to; it would be up to the individual deploying the software.

### Justification

Please see above

AgentTNT:
Break out the domain page into a separate Rendering.  This is the first pass at this.  It is not complete yet. Looking to coninue to break out the domain page into smaller components from Jinja2 template variables into JSON objects.
This will allow for easier client side rendering of the domain page.


AgentTNT:
Changed behavior of modal of the apply changes to be persistent for user experience so they are aware of the entire workflow and are aware of page reload after successful submission of changes

AgentTNT:
- Breakout record helper modals to own page and import them to other pages, to create a single source of modals for records moving forward

osos:
Is this still needed to complete Milestone 0.4.3 ?
Or could this be postponed to 0.5.0 ?

Issue #1607, opened 2023-07-01
Provide support for catalog zones


feldsam:
### PDA version

0.4.1

### Feature type

New functionality

### Proposed functionality

I tested catalog zone with pdns 4.7.3 and dev branch of pda. Figured out, that only few things should be added to pda app:

- when creating new zone:
  - additional zone type `PRODUCER` - identiffies catalog zone - probably not needed here, can be created one time using CLI
  - add new dropdown to select catalog zone - show only if there is any
  - add new dropdown to select tsig key to use - to use secure transfers

That is all for minimal implementation. On dashboard, pda is properly showing type `producer` for catalog zone, so no more work is needed.

I can take care of it and submit pull request.

### Use case

I like to use it with non-pdns slave server, I want to use bind for slave. DNS system will be more resilent, when I use different software.

Catalog Zones is a feature allowing easy provisioning of zones to secondary servers. A "catalog zone" is a special DNS zone that contains a list of other zones to be served, along with their configuration parameters.  The zones listed in a catalog zone are called "member zones".  When a catalog zone is loaded or transferred to a secondary server that supports this functionality, the secondary server creates the member zones automatically.  When the catalog zone is updated (for example, to add or delete member zones, or change their configuration parameters) those changes are immediately put into effect. Because the catalog zone is a normal DNS zone, these configuration changes can be propagated using the standard AXFR/IXFR zone transfer mechanism.

### Database changes

none, type field is varchar(8) so it fits

### External dependencies

none

AzorianMatt:
@feldsam splendid work on this! I appreciate the efforts and it sounds like you did your homework.

One of our contribution team members is currently working on some changes related to The zone editor, so I want to involve him in the conversation as well as there will be impacts both ways.

@AgentTNT thoughts?

martijngoorman:
I just commented on https://github.com/PowerDNS-Admin/PowerDNS-Admin/issues/1348 for a status, but this issue looks prommising.
Maybe add @Metrax to this issue?

Metrax:
There is a little bit more todo than the points at "Proposed functionality"
Catalog Zones should be only available for PDNS Versions > 4.7, we need to introduce a new version detection system here, else we will get too much support requests here.
Also since 4.7.4 there is a function "list-member-zones" which we should implement.

feldsam:
[...]

PR #1641, opened 2023-08-07
Provide Catalog Zone Support

feldsam:
REFS #1607


AzorianMatt:
@feldsam this morning I addressed a bug in PDA based on the use of zone names with forward slashes in the name. Would you mind applying this quick fix to your changes to the domain.py model file please? Reference PR 1707 for a direct example of the other changes I made which are very quick and simple.

feldsam:
Hello @AzorianMatt , why I should apply your PR to my? When your will be merged before mine, them I solve possible conflicts.

AzorianMatt:
@feldsam I was asking you to apply the simple URL encoding change so that I don't have to open a separate PR or do a separate set of commits to correct yours once merged. Does that make more sense?

Also, how should I test this? I have your copy checked out, but none of my record sets are loading when opening a zone to edit. Not really sure what is wrong there.

feldsam:
No, it doesn't make sense. Why you do not merge your PR before mine? Them I pull changes and apply to my branch, slve conflicts.

AzorianMatt:
@feldsam I don't think you understand, I already merged the changes into `dev`. You can merge them now if you like, but what I am asking is for you to replicate the type of change I made into your new code that is part of this PR. The code you added, will essentially re-create the bug I just fixed because it also doesn't account for the issue I was correcting. Just review the PR and you should see what I mean.

PR #1642, opened 2023-08-07
Merge zone editor record action controls into single column

feldsam:
This PR enhances domains table on dashboard. Action buttons, each in separate column consumes lot of space on smaller screens. This PR joins action buttons into single column.

### DIFF
--- a/powerdnsadmin/templates/domain.html
+++ b/powerdnsadmin/templates/domain.html
@@ -86,11 +86,7 @@
                                     <th>Data</th>
                                     {% if domain.type != 'Slave' %}
                                         <th>Comment</th>
-                                        <th>Edit</th>
-                                        <th>Delete</th>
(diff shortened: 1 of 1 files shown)


AzorianMatt:
@feldsam I'm confused by this PR. I don't recall it being discussed and approved (missing an issue reference). Furthermore, the current release of PDA doesn't have multiple columns for action buttons, but instead a single item that generates a drop-down menu of options.

What am I missing?

feldsam:
Hi, sorry, wrong description, it is actually in domain detail - list of records. Check screenshot. In actual version, each button is in separate table column

Screenshot 2023-11-24 at 16 54 08

AzorianMatt:
@feldsam ahh okay, that makes more sense. Give me awhile to get caught up with everything else and I'll put this on the list for review. Thanks!

PR #1656, opened 2023-08-29
Correct OIDC logout to properly support Keycloak

teunis90:
Keycloak expects these two values:

post_logout_redirect_uri (and not redirect_uri)
id_token_hint (which is recommended by the spec: https://openid.net/specs/openid-connect-rpinitiated-1_0.html, but required by keycloak)

Tested on Keycloak version: 20.0.5

Fixes https://github.com/PowerDNS-Admin/PowerDNS-Admin/issues/1222

### DIFF
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -652,10 +652,11 @@ def logout():
 
     redirect_uri = url_for('index.login')
     oidc_logout = Setting().get('oidc_oauth_logout_url')
+    id_token_hint = session.get('oidc_token').get('id_token', None)
 
-    if 'oidc_token' in session and oidc_logout:
-        redirect_uri = "{}?redirect_uri={}".format(
-            oidc_logout, url_for('index.login', _external=True))
+    if id_token_hint and oidc_logout:
+        redirect_uri = "{}?post_logout_redirect_uri={}&id_token_hint={}".format(
+            url_for('index.login', _external=True), oidc_logout, id_token_hint)
 
     # Clean cookies and flask session
     clear_session()
--- a/powerdnsadmin/routes/index.py
+++ b/powerdnsadmin/routes/index.py
@@ -652,7 +652,7 @@ def logout():
 
     redirect_uri = url_for('index.login')
     oidc_logout = Setting().get('oidc_oauth_logout_url')
-    id_token_hint = session.get('oidc_token').get('id_token', None)
+    id_token_hint = session.get('oidc_token').get('id_token')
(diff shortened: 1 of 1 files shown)


powerdnsadmin/routes/index.py
```diff
@@ -652,10 +652,11 @@ def logout():
 
     redirect_uri = url_for('index.login')
     oidc_logout = Setting().get('oidc_oauth_logout_url')
+    id_token_hint = session.get('oidc_token').get('id_token')
```
AgentTNT, line 655: If Local DB access is still enabled and a user goes to logout, they will hit an error as this value is not defined for their session

powerdnsadmin/routes/index.py
```diff
@@ -652,10 +652,11 @@ def logout():
...
 
-    if 'oidc_token' in session and oidc_logout:
-        redirect_uri = "{}?redirect_uri={}".format(
-            oidc_logout, url_for('index.login', _external=True))
+    if id_token_hint and oidc_logout:
+        redirect_uri = "{}?post_logout_redirect_uri={}&id_token_hint={}".format(
+            url_for('index.login', _external=True), oidc_logout, id_token_hint)
```
AgentTNT, line 660: `    if 'oidc_token' in session and oidc_logout:
        id_token_hint = session.get('oidc_token').get('id_token')
        redirect_uri = "{}?post_logout_redirect_uri={}&id_token_hint={}".format(
            oidc_logout, url_for('index.login', _external=True), id_token_hint)
`

This makes powerdns-admin redirect to the keycloak or other OIDC logout endpoint first, then get redirected back to the login page, fully logging out all OIDC session cookies within the browser whereas before the IDP was still staying signed in and not allowing users to login with new creds when clicking the OIDC login

powerdnsadmin/routes/index.py
teunis90: @AgentTNT, can you reproduce the not-defined error? What do you suggest?

powerdnsadmin/routes/index.py
teunis90: I'll update the PR with your suggestion.

AzorianMatt: @teunis90 the time is nearing quickly for me to make the final version release of this edition of PDA. If you would like to see this make it in, I would recommend getting that last change request implemented very soon. Thanks!

keliansb:
Can this PR be reopened? We are facing the same issue with Keycloak.

Issue #1657, opened 2023-08-31
Local user setup utilizes case-sensitive verification of username and email


ronanac3:
### PDA version

0.4.1

### Python version

3.7

### Steps to Reproduce

1. In PowerDNS Web Interface, goto Application -> Users
2. For a given test user, edit their email to case usage e.g first.last@email.com.au -> First.Last@email.com.au
3. Now an additional user can be created with the email first.last@email.com.au (following the example above)

This occurred for me when moving our PowerDNS-Admin to a Kubernetes pod. In this process, Azure OAuth was set up which auto creates accounts for users using our new company email domain as their username. However, although the new users had new usernames, many users had the same email. Of these users, many had the same case email (thus a new account could not be created as their email is already in use) and many had the same email but with a different case (e.g. first.last@email.com.au -> First.Last@email.com.au, thus allowing them to have a new account created). 

### Expected Behavior

It is expected that multiple user accounts cannot have the same email with different casing. The check for an existing email (and username too?) should be case insensitive. This should be able to be solved simply using .lower() on the strings when comparing the two.

### Observed Behavior

Accounts are allowed to be created with duplicate usernames and emails if the string casing is not equal. Thus, a user with an email equal to first.last@email.com.au can coexist with a user having an email equal to First.Last@email.com.au despite the emails being the same.

ronanac3:
Additionally, can we/I replace 
`current_app.logger.warning('Unable to create ' + azure_username)` 
with  
`current_app.logger.warning('Unable to create ' + azure_username + 'Reasoning:' + result['msg'])`

To add further information when debugging.

AzorianMatt:
This has been resolved by PR 1658.

PR #1658, opened 2023-08-31
Fixes local user setup to perform case-insensitive verification of existing usernames / emails

ronanac3:
### Fixes: #1657 

I have updated the create_local_user() function to be case sensitive when checking the database for existing users with the same username and/or email. Additionally, when this function fails it returns its result status and message, as this message is useful for debugging, I have added it to the log message.

### DIFF
--- a/powerdnsadmin/models/user.py
+++ b/powerdnsadmin/models/user.py
@@ -408,7 +408,7 @@ class User(db.Model):
         Create local user witch stores username / password in the DB
         """
         # check if username existed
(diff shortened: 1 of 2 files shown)


AzorianMatt:
@ronanac3 please be more cautious with future submissions. I had to make a change in `dev` following this PR merge because it broke new user registration as it turns out. Here is what the fix actually looks like now;

`user = User.query.filter(str(User.username).lower() == self.username.lower()).first()`
`user = User.query.filter(str(User.email).lower() == self.email.lower()).first()`

Issue #1686, opened 2023-11-02
Use of the `before_app_first_request` event removed in Flask 2.3


pbsds:
### PDA version

0.4.1

### Python version

3.11

### Steps to Reproduce

Build the yarn assets with updated python deps

### Expected Behavior

It builds

### Observed Behavior

```
  File "/usr/lib/python3.11/site-packages/flask/cli.py", line 64, in find_best_app
    app = app_factory()
          ^^^^^^^^^^^^^
  File "/build/source/deps/powerdns-admin-assets/powerdnsadmin/__init__.py", line 12, in create_app
    from . import models, routes, services
  File "/build/source/deps/powerdns-admin-assets/powerdnsadmin/routes/__init__.py", line 6, in <module>
    from .index import index_bp
  File "/build/source/deps/powerdns-admin-assets/powerdnsadmin/routes/index.py", line 49, in <module>
    @index_bp.before_app_first_request
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Blueprint' object has no attribute 'before_app_first_request'. Did you mean: 'before_app_request'?
```

pbsds:
I used flask 2.3.3, context: https://github.com/pallets/flask/issues/4605

AzorianMatt:
This issue should be resolved by PR 1705 and is now present in the `dev` branch.

AzorianMatt:
This was a bad and rushed solution on my part which created new issues. I had to revert the change.

PR #1705, opened 2023-11-24
Update index router to replace the use of the deprecated `before_app_first_request` event with `record_once`

AzorianMatt:
### Fixes: #1686

I updated the `powerdnsadmin/routes/index.py` router method `register_modules` to use the `Blueprint.record_once` event in-place of the deprecated `Blueprint.before_app_first_request` event.


AzorianMatt:
This was a bad "fix" on my part. Reverting.

PR #1747, opened 2024-02-01
Implement catalog zone support

### LINKED ISSUE #1607: Provide support for catalog zones

AzorianMatt:
Resolves #1607

Implements catalog zone support, provided by @feldsam.


PAStheLoD:
Hello,

Excuse me for going against Matt's the implied intent (of closing/pausing issue reporting).

I am reporting this (because the fix seems easy) and it might/hopefully help others.

So blindly "upgrading" by using the `dev`[0] docker image tag (instead of nothing, which defaults to `latest`[1]) leads to a HTTP 500 when creating a new zone.

```python
[2024-08-16 11:38:41 +0000] [9] [DEBUG] POST /domain/add
[2024-08-16 11:38:41 +0000] [9] [ERROR] Cannot add zone. Error: list index out of range
[2024-08-16 11:38:41,189] [domain.py:460] ERROR - Cannot add zone. Error: list index out of range
[2024-08-16 11:38:41 +0000] [9] [DEBUG] Traceback (most recent call last):
  File "/app/powerdnsadmin/routes/domain.py", line 309, in add
    catalog_name = request.form.getlist('catalog_name')[0]
IndexError: list index out of range
```

This is because the new feature is not enabled by default, so "catalog_name" is not present in the HTTP POST form data and so there's an empty list, hence the indexing error.

The solution is to create one catalog zone:

```
pdnsutil create-zone catalog.internal
pdnsutil set-kind catalog.internal producer
```


Thanks!



[0] powerdnsadmin/pda-legacy   dev       sha256:556da73a8fd1fa9c33ada3c366eba15bc316dd9eea21ef075305072dee0a05a2   17c4c0057308   6 months ago    218MB

[1] powerdnsadmin/pda-legacy   latest    sha256:0fc92c8ef63a6704083ca1989c83d2b747f7abc0e5045d56d9c69a76e08a9a5e   c20024deaf49   6 months ago    218MB

PR #1764, opened 2024-03-12
Update oauth.md

C0mmanderCrud:
Correcting a wording change in the documentation that stops OAUTH working when intergrating with entraID as mail will throw a error as scope setup is looking for email


AzorianMatt:
@C0mmanderCrud please try reading both the posted announcements and the contribution guidelines. This will be rejected and closed.

PR #1797, opened 2024-10-30
Remove leading slash

John222:
Remove the leading slash to let dns gui connect to powerdns servers which are behind a proxy under subpath.

Fixes the issue to connect to PowerDNS instances behind a proxy which lives under a subpath.

The old code behaves like this when you enter a url:

Example 1 (breaks the url):

`
urljoin('https://dns.example.com/pdns-authoritative', '/api/v1' + '/servers/localhost/zones')
'https://dns.example.com/api/v1/servers/localhost/zones'
`
Example 2 (breaks the url):
`
urljoin('https://dns.example.com/pdns-authoritative/', '/api/v1' + '/servers/localhost/zones')
'https://dns.example.com/api/v1/servers/localhost/zones'
`

The proposed change fixes it

Example 1 (keeps the url):

`
urljoin('https://dns.example.com/pdns-authoritative/', 'api/v1' + '/servers/localhost/zones')
'https://dns.example.com/pdns-authoritative/api/v1/servers/localhost/zones'
`

In case some entered the url without a trailing slash, it would break existing deployed applications.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -67,7 +67,7 @@ jobs:
           context: ./
           file: ./docker/Dockerfile
           push: true
-          tags: powerdnsadmin/pda-legacy:${{ github.ref_name }}
+          tags: john0222/pda-legacy:${{ github.ref_name }}
 
       - name: Docker Image Release Tagging
         uses: docker/build-push-action@v4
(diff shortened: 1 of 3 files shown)


DaanSelen:
bump

AgentTNT:
Given delta, this has been merged to local dev, will be pushed to public dev shortly.

PR #1803, opened 2024-12-12
escaping zone name in zones api url to fully support RFC2317

metrax:
PDA gets back error 405 from PDNS API when trying to open a RFC2317 zone. The problem is,  that PDA is not escaping the zone in the url at all points. This PR fixed this issue.

Example:
`[record.py:386] ERROR - Cannot apply record changes to zone 0/28.13.xxx.xxx.in-addr.arpa. Error: Error while fetching http://[::1]:8081/api/v1/servers/localhost/zones/0/28.13.xxx.xxx.in-addr.arpa. Returned status 405 and content Method Not Allowed
`


DaanSelen:
bump

PR #1868, opened 2026-08-03
Fix URL encoding for PowerDNS zone names

metrax:
This updates PowerDNS API requests that include a zone name in the URL path.

Zone names are now encoded with quote_plus() before being added to the request URL. This covers domain, record, DNSSEC, AXFR and API zone-deletion requests.

The existing urllib.parse usage was also consolidated into module-level imports.

I was unable to open an issue for this change because issue creation is currently unavailable in the repository.


AgentTNT:
Thanks for the PR - Issues are now enabled again for the repo

PR #1889, opened 2026-08-09
Fix OIDC update_token callback signature and implement safe session token merge

jhass:
Authlib calls the `update_token` callback with extra keyword arguments (`refresh_token`, `access_token`) that identify which old token was consumed during a refresh. The existing single-argument signature caused a hard runtime failure during any OIDC token refresh:

```
oidc_oauth.<locals>.update_token() got an unexpected keyword argument 'refresh_token'
```

### `powerdnsadmin/services/oidc.py`
- Changed `update_token(token)` → `update_token(token, **kwargs)` to accept Authlib's extra kwargs without raising.
- Replaced the naive `session['oidc_token'] = token` assignment with a three-step merge:
  1. **Base**: existing session token (preserves stable fields like `refresh_token`)
  2. **Overlay**: incoming `token` mapping (new `access_token`/`expires_at` always win)
  3. **Fallback**: kwarg-provided fields only when still absent — handles IdPs that omit `refresh_token` on silent refresh responses (RFC 6749 §6 permits reuse of the consumed token)
- Non-mapping values in either session or incoming token are silently ignored.

### `tests/unit/test_oidc_service.py`
- Added 7 tests covering: kwarg acceptance, kwarg-vs-token precedence, session preservation of `refresh_token`, stale `access_token` overwrite, and defensive handling of garbage session/token values.

### DIFF
--- a/powerdnsadmin/services/oidc.py
+++ b/powerdnsadmin/services/oidc.py
@@ -33,9 +33,41 @@ def oidc_oauth():
     def fetch_oidc_token():
         return session.get('oidc_token')
 
-    def update_token(token):
-        session['oidc_token'] = token
-        return token
+    def update_token(token, **kwargs):
+        # Authlib may call update_token with extra keyword arguments such as
+        # `refresh_token` or `access_token` that were used to obtain the new
+        # token.  Accept them gracefully so the callback never raises
+        # "unexpected keyword argument".
+        #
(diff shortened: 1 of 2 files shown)


AgentTNT:
Thank you for the PR! About to roll this into the larger auth re-work as well :)

